Issue #1, opened 2022-09-03
Anyone there?


olanod:
In case there's no conversation going on already somewhere else, consider this a conversation starter.  
e.g.
- `wasi-http` should ...
- `wasi-http` should not ...

linclark:
This is currently waiting on the async support that is being actively discussed in the WASI meetings. We won't be able to make progress on this until that's further along, so I'm going to close this one out.

PR #3, opened 2023-02-17
add Wit interfaces

lukewagner:
This PR proposes some initial Wit interfaces and a `wasi:http/proxy` world using them that's mostly a synthesis of a lot of the existing proposals.  As the proposal develops, I expect us to add new "bigger" worlds than `wasi:http/proxy` world (either here or in other proposals), but this `wasi:http/proxy` world seems like perhaps a good minimal concrete starting point.  There's still a lot more to fill in to make this a proper WASI proposal, but perhaps this is enough for us to chew on now.  Lastly, I still consider this very much in a draft state so happy to iterate with folks in this PR.

(cc @PiotrSikora @brendandburns @eduardomourar)

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ This proposal currently only contains the proposed Wit interfaces with light
 explanation in comments; more work is necessary to fully document the proposal.
 The Wit comments annotate where the proposed interface is expected to change in
 the short term (for Preview2) once resources and handles are re-added to Wit,
-and then after that (for Preview2) once native stream support is added to the
+and then after that (for Preview3) once native stream support is added to the
 Component Model and Wit.
 
 The `wit` directory currently validates and can generate bindings with:
--- a/README.md
+++ b/README.md
@@ -1,3 +1,38 @@
 # WASI HTTP
 
-(This is a placeholder so there can be a PR to fill in the contents.)
+A proposed [WebAssembly System Interface](https://github.com/WebAssembly/WASI) API.
+
+This proposal currently only contains the proposed Wit interfaces with light
+explanation in comments; more work is necessary to fully document the proposal.
+The Wit comments annotate where the proposed interface is expected to change in
+the short term (for Preview2) once resources and handles are re-added to Wit,
+and then after that (for Preview2) once native stream support is added to the
+Component Model and Wit.
+
+The `wit` directory currently validates and can generate bindings with:
+```
+wit-bindgen c wit/ --world proxy
+```
+or can be manipulated in other ways with:
+```
+wasm-tools component wit wit/ ...
+```
+
+The HTTP proposal depends on the WASI IO and Logging proposals. For simplicity,
+the Wit files for these proposals are currently copied into the `wit/deps`
+directory and will be updated periodically to match their respective proposals.
+As the Wit tooling develops, we should be able to avoid this form of manual
+vendoring.
+
+### Current Phase
+
+wasi-http is currently in [Phase 1](https://github.com/WebAssembly/WASI/blob/main/Proposals.md).
+
+### Champions
+
+Piotr Sikora, Jiaxiao Zhou, Dan Chiarlone, David Justice
+
+### TODO
+
+This readme needs to be expanded to cover a number of additional fields suggested in the
+[WASI Proposal template](https://github.com/WebAssembly/wasi-proposal-template).
--- a/wit/deps/logging/handler.wit
+++ b/wit/deps/logging/handler.wit
@@ -1,0 +1,31 @@
+/// # WASI Logging API
+///
+/// WASI Logging is a logging API intended to let users emit log messages with
+/// simple priority levels and context values.
+default interface handler {
+  /// A log level, describing a kind of message.
+  enum level {
+     /// Describes messages about the values of variables and the flow of control
+     /// within a program.
+     trace,
+
+     /// Describes messages likely to be of interest to someone debugging a program.
+     debug,
+
+     /// Describes messages likely to be of interest to someone monitoring a program.
+     info,
+
+     /// Describes messages indicating hazardous situations.
+     warn,
+
+     /// Describes messages indicating serious errors.
+     error,
+  }
+
+  /// Emit a log message.
+  ///
+  /// A log message has a `level` describing what kind of message is being sent,
+  /// a context, which is an uninterpreted string meant to help consumers group
+  /// similar messages, and a string containing the message text.
+  log: func(level: level, context: string, message: string)
+}
--- a/wit/deps/logging/handler.wit
+++ b/wit/deps/logging/handler.wit
@@ -1,31 +1,32 @@
-/// # WASI Logging API
-///
 /// WASI Logging is a logging API intended to let users emit log messages with
 /// simple priority levels and context values.
 default interface handler {
-  /// A log level, describing a kind of message.
-  enum level {
-     /// Describes messages about the values of variables and the flow of control
-     /// within a program.
-     trace,
+    /// A log level, describing a kind of message.
+    enum level {
+       /// Describes messages about the values of variables and the flow of
+       /// control within a program.
+       trace,
 
-     /// Describes messages likely to be of interest to someone debugging a program.
-     debug,
+       /// Describes messages likely to be of interest to someone debugging a
+       /// program.
+       debug,
 
-     /// Describes messages likely to be of interest to someone monitoring a program.
-     info,
+       /// Describes messages likely to be of interest to someone monitoring a
+       /// program.
+       info,
 
-     /// Describes messages indicating hazardous situations.
-     warn,
+       /// Describes messages indicating hazardous situations.
+       warn,
 
-     /// Describes messages indicating serious errors.
-     error,
-  }
+       /// Describes messages indicating serious errors.
+       error,
+    }
 
-  /// Emit a log message.
-  ///
-  /// A log message has a `level` describing what kind of message is being sent,
-  /// a context, which is an uninterpreted string meant to help consumers group
-  /// similar messages, and a string containing the message text.
-  log: func(level: level, context: string, message: string)
+    /// Emit a log message.
+    ///
+    /// A log message has a `level` describing what kind of message is being
+    /// sent, a context, which is an uninterpreted string meant to help
+    /// consumers group similar messages, and a string containing the message
+    /// text.
+    log: func(level: level, context: string, message: string)
 }
--- a/wit/incoming-handler.wit
+++ b/wit/incoming-handler.wit
@@ -7,7 +7,7 @@
 //   that takes a `request` parameter and returns a `response` result.
 //
 default interface incoming-handler {
-  use pkg.types.{incoming-request, request-options, response-outparam}
+  use pkg.types.{incoming-request, response-outparam}
 
   // The `handle` function takes an outparam instead of returning its response
   // so that the component may stream its response while streaming any other
@@ -19,7 +19,6 @@ default interface incoming-handler {
   // its success or failure.
   handle: func(
     request: incoming-request,
-    response-out: response-outparam,
-    options: option<request-options>
+    response-out: response-outparam
   )
 }
--- a/wit/incoming-handler.wit
+++ b/wit/incoming-handler.wit
@@ -7,7 +7,7 @@
 //   that takes a `request` parameter and returns a `response` result.
 //
 default interface incoming-handler {
-  use pkg.types.{incoming-request, response-outparam}
+  use pkg.types.{incoming-request, request-options, response-outparam}
 
   // The `handle` function takes an outparam instead of returning its response
   // so that the component may stream its response while streaming any other
@@ -17,5 +17,9 @@ default interface incoming-handler {
   // output stream. While this post-response execution is taken off the
   // critical path, since there is no return value, there is no way to report
   // its success or failure.
-  handle: func(request: incoming-request, response-out: response-outparam)
+  handle: func(
+    request: incoming-request,
+    response-out: response-outparam,
+    options: option<request-options>
+  )
 }
--- a/wit/incoming-handler.wit
+++ b/wit/incoming-handler.wit
@@ -1,0 +1,21 @@
+// The `wasi:http/incoming-handler` interface is meant to be exported by
+// components and called by the host in response to a new incoming HTTP
+// response.
+//
+//   NOTE: in Preview3, this interface will be merged with
+//   `wasi:http/outgoing-handler` into a single `wasi:http/handler` interface
+//   that takes a `request` parameter and returns a `response` result.
+//
+default interface incoming-handler {
+  use pkg.types.{incoming-request, response-outparam}
+
+  // The `handle` function takes an outparam instead of returning its response
+  // so that the component may stream its response while streaming any other
+  // request or response bodies. The callee MUST write a response to the
+  // `response-out` and then finish the response before returning. The `handle`
+  // function is allowed to continue execution after finishing the response's
+  // output stream. While this post-response execution is taken off the
+  // critical path, since there is no return value, there is no way to report
+  // its success or failure.
+  handle: func(request: incoming-request, response-out: response-outparam)
+}
--- a/wit/outgoing-handler.wit
+++ b/wit/outgoing-handler.wit
@@ -1,0 +1,15 @@
+// The `wasi:http/outgoing-handler` interface is meant to be imported by
+// components and implemented by the host.
+//
+//   NOTE: in Preview3, this interface will be merged with
+//   `wasi:http/outgoing-handler` into a single `wasi:http/handler` interface
+//   that takes a `request` parameter and returns a `response` result.
+//
+default interface outgoing-handler {
+  use pkg.types.{outgoing-request, incoming-response, error}
+
+  // The parameter and result types of the `handle` function allow the caller
+  // to concurrently stream the bodies of the outgoing request and the incoming
+  // response.
+  handle: func(request: outgoing-request) -> result<incoming-response, error>
+}
--- a/wit/outgoing-handler.wit
+++ b/wit/outgoing-handler.wit
@@ -6,10 +6,13 @@
 //   that takes a `request` parameter and returns a `response` result.
 //
 default interface outgoing-handler {
-  use pkg.types.{outgoing-request, future-incoming-response}
+  use pkg.types.{outgoing-request, request-options, future-incoming-response}
 
   // The parameter and result types of the `handle` function allow the caller
   // to concurrently stream the bodies of the outgoing request and the incoming
   // response.
-  handle: func(request: outgoing-request) -> future-incoming-response
+  handle: func(
+    request: outgoing-request,
+    options: option<request-options>
+  ) -> future-incoming-response
 }
--- a/wit/outgoing-handler.wit
+++ b/wit/outgoing-handler.wit
@@ -6,10 +6,10 @@
 //   that takes a `request` parameter and returns a `response` result.
 //
 default interface outgoing-handler {
-  use pkg.types.{outgoing-request, incoming-response, error}
+  use pkg.types.{outgoing-request, future-incoming-response}
 
   // The parameter and result types of the `handle` function allow the caller
   // to concurrently stream the bodies of the outgoing request and the incoming
   // response.
-  handle: func(request: outgoing-request) -> result<incoming-response, error>
+  handle: func(request: outgoing-request) -> future-incoming-response
 }
--- a/wit/proxy.wit
+++ b/wit/proxy.wit
@@ -1,0 +1,32 @@
+// The `wasi:http/proxy` world captures a widely-implementable intersection of
+// hosts that includes HTTP forward and reverse proxies. Components targeting
+// this world may concurrently stream in and out any number of incoming and
+// outgoing HTTP requests.
+default world proxy {
+
+  // This is the default logging handler to use when user code simply wants to
+  // log to a developer-facing console (e.g., via `console.log()`).
+  import console: logging.handler
+
+  // TODO: add `import metrics: metrics.counters`
+
+  // This is the default handler to use when user code simply wants to make an
+  // HTTP request (e.g., via `fetch()`) but doesn't otherwise specify a
+  // particular handler.
+  import default-upstream-HTTP: pkg.outgoing-handler
+
+  // TODO: Once the underlying Wit template machinery is implemented, add:
+  //
+  //  import upstreams: interface {
+  //    *: pkg.outgoing-handler
+  //  }
+  //
+  // which will allow a component to import any number of non-default backends
+  // that HTTP requests can be dispatched to.
+
+  // The host delivers incoming HTTP requests to a component by calling the
+  // `handle` function of this exported interface. A host may arbitrarily reuse
+  // or not reuse component instance when delivering incoming HTTP requests and
+  // thus a component must be able to handle 0..N calls to `handle`.
+  export HTTP: pkg.incoming-handler
+}
--- a/wit/proxy.wit
+++ b/wit/proxy.wit
@@ -26,7 +26,7 @@ default world proxy {
   // This is the default handler to use when user code simply wants to make an
   // HTTP request (e.g., via `fetch()`) but doesn't otherwise specify a
   // particular handler.
-  import default-upstream-HTTP: pkg.outgoing-handler
+  import default-outgoing-HTTP: pkg.outgoing-handler
 
   // TODO: Once the underlying Wit template machinery is implemented, add:
   //
--- a/wit/proxy.wit
+++ b/wit/proxy.wit
@@ -3,10 +3,23 @@
 // this world may concurrently stream in and out any number of incoming and
 // outgoing HTTP requests.
 default world proxy {
+  // HTTP proxies have access to time and randomness.
+  import random: random.random
+  // TODO: add `import wall-clock: clocks.wall-clock`
+  // TODO: add `import monotonic-clock: clocks.monotonic-clock`
 
   // This is the default logging handler to use when user code simply wants to
   // log to a developer-facing console (e.g., via `console.log()`).
   import console: logging.handler
+
+  // TODO: Once the underlying Wit template machinery is implemented, add:
+  //
+  //   import loggers: interface {
+  //     *: logging.handler
+  //   }
+  //
+  // which will allow a component to import any number of non-default logging
+  // backends that different categories of log messages can be sent to.
 
   // TODO: add `import metrics: metrics.counters`
 
--- a/wit/types.wit
+++ b/wit/types.wit
@@ -53,29 +53,26 @@ default interface types {
   type headers = fields
   type trailers = fields
 
-  // The following block defines the `body` type which corresponds to the HTTP
-  // standard Contents. With Preview3, all of these fields can be replaced by a
-  // single type definition:
-  //
-  //   type body = stream<u8, option<trailers>>
-  //
-  // In the interim, we need to use separate `input-stream` and `output-stream`
-  // resource types defined by `wasi:io/streams`. The `finish-` functions
-  // emulate the stream's result value and MUST be called exactly once after
-  // the final read/write from/to the stream before dropping the stream.
-  type incoming-body = input-stream
-  type outgoing-body = output-stream
-  finish-incoming-body: func(body: incoming-body) -> option<trailers>
-  finish-outgoing-body: func(body: outgoing-body, trailers: option<trailers>)
+  // The following block defines stream types which corresponds to the HTTP
+  // standard Contents and Trailers. With Preview3, all of these fields can be
+  // replaced by a stream<u8, option<trailers>>. In the interim, we need to
+  // build on separate resource types defined by `wasi:io/streams`. The
+  // `finish-` functions emulate the stream's result value and MUST be called
+  // exactly once after the final read/write from/to the stream before dropping
+  // the stream.
+  type incoming-stream = input-stream
+  type outgoing-stream = output-stream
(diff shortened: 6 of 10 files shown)


Mossaka (approved):
Hey Luke, thanks for the PR! This looks great to me!

wit/incoming-handler.wit
```diff
@@ -0,0 +1,21 @@
+// The `wasi:http/incoming-handler` interface is meant to be exported by
+// components and called by the host in response to a new incoming HTTP
+// response.
+//
+//   NOTE: in Preview3, this interface will be merged with
+//   `wasi:http/outgoing-handler` into a single `wasi:http/handler` interface
+//   that takes a `request` parameter and returns a `response` result.
+//
+default interface incoming-handler {
```
brendandburns, line 9: Do we want to split this out into separate PRs? I feel like it's going to be a heavy lift to do both implementations in a single PR.

Of course, if it's ok to only have a partial implementation, that works too.

brendandburns:
This looks great for a start. I had a couple of questions:

The first is that is it ok to do a partial implementation of the spec when we start adding it to wasmtime? If not I worry that this is so comprehensive it will be a large change to get landed.

The second is that I'm expecting this API will adapt and change as we (and others) gain experience with it. Thus my review is "is it good enough?" not "is it perfect?" I hope that works.

lukewagner:
@brendandburns Thanks for the review and good questions.

> The first is that is it ok to do a partial implementation of the spec when we start adding it to wasmtime?

I think it's pretty normal for Wasmtime to land implementations incrementally and so adding a partial implementation makes sense to me, but I'd be interested to hear from @pchickey or @alexcrichton.  It might be useful to open an early issue to discuss the rough shape of the end-state design, though, since I think by now we have a few distinct Wasmtime embeddings that are pretty interested in reusing a common shared HTTP implementation.

> Thus my review is "is it good enough?" not "is it perfect?" I hope that works.

Yes, that is totally the appropriate level of review at this early stage.  Iterating on the design in response to implementation feedback is an essential and expected part of both the [core wasm](https://github.com/WebAssembly/meetings/blob/main/process/phases.md) and [WASI](https://github.com/WebAssembly/WASI/blob/main/Contributing.md) phases process.

alexcrichton:
Yes landing incrementally in Wasmtime is ok. The main threshold for landing an experimental feature in Wasmtime is "it doesn't affect anything else or otherwise has consensus about how it should be designed". The consensus part is nontrivial to acquire but I suspect an implementation here wouldn't affect other pieces so consensus shouldn't be necessary.

wit/proxy.wit
```diff
@@ -0,0 +1,32 @@
...
+  //  }
+  //
+  // which will allow a component to import any number of non-default backends
+  // that HTTP requests can be dispatched to.
+
+  // The host delivers incoming HTTP requests to a component by calling the
+  // `handle` function of this exported interface. A host may arbitrarily reuse
+  // or not reuse component instance when delivering incoming HTTP requests and
+  // thus a component must be able to handle 0..N calls to `handle`.
+  export HTTP: pkg.incoming-handler
```
eduardomourar, line 31:
```suggestion
  export http: pkg.incoming-handler
```

wit/types.wit
```diff
@@ -0,0 +1,128 @@
+// The `wasi:http/types` interface is meant to be imported by components to
+// define the HTTP resource types and operations used by the component's
+// imported and exported interfaces.
+default interface types {
+  
+  // This type corresponds to HTTP standard Methods.
+  variant method {
```
sdeleuze, line 7: Should an additional value taking a `string` parameter, like the `other(string)` value in `scheme`, be added in order to support [extension methods](https://datatracker.ietf.org/doc/html/rfc2616#section-5.1.1) used for example in [WebDAV](http://www.webdav.org/specs/rfc4918.html#http.methods.for.distributed.authoring)?

wit/proxy.wit
```diff
@@ -0,0 +1,32 @@
...
+  // This is the default logging handler to use when user code simply wants to
+  // log to a developer-facing console (e.g., via `console.log()`).
+  import console: logging.handler
+
+  // TODO: add `import metrics: metrics.counters`
+
+  // This is the default handler to use when user code simply wants to make an
+  // HTTP request (e.g., via `fetch()`) but doesn't otherwise specify a
+  // particular handler.
+  import default-upstream-HTTP: pkg.outgoing-handler
```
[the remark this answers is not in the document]
lukewagner: The intention with using an all-caps [`word`](https://github.com/WebAssembly/component-model/blob/main/design/mvp/Explainer.md?plain=1#L210) here is to indicate that "HTTP" is an acronym, which lets the language bindings to use casing appropriate for acronyms (as described in more detail [here](https://github.com/WebAssembly/component-model/blob/main/design/mvp/Explainer.md?plain=1#L245-L253)).

wit/outgoing-handler.wit
```diff
@@ -0,0 +1,15 @@
...
+//   `wasi:http/outgoing-handler` into a single `wasi:http/handler` interface
+//   that takes a `request` parameter and returns a `response` result.
+//
+default interface outgoing-handler {
+  use pkg.types.{outgoing-request, incoming-response, error}
+
+  // The parameter and result types of the `handle` function allow the caller
+  // to concurrently stream the bodies of the outgoing request and the incoming
+  // response.
+  handle: func(request: outgoing-request) -> result<incoming-response, error>
```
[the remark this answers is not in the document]
lukewagner: Yep!  That does raise a really good point about something I may have overlooked here: for Preview2, should this function return a resource that can be passed to `poll_oneoff` (just like the streams), thereby allowing concurrent execution before the `result` is known?  IIUC, knowing whether to return the success or failure case requires a network round trip which may result in a timeout so it can take a "long" time, so it seems like "yes", we do need to emulate a `future` here.  Is that right, though?

wit/types.wit
lukewagner: Ah, good point!  [Added](https://github.com/WebAssembly/wasi-http/pull/3/commits/67f6b2f016524a407026a68ddfd218d879aefa0f)

wit/outgoing-handler.wit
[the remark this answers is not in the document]
eduardomourar: Yes, I would believe so. To me it seems that we need to take this into account because there is at least one scenario that a `future` applies.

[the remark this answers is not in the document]
eduardomourar: I guess we could get away with not worrying about this as long as the different languages' codegen can convert an implementation that returns `future` to one that does not internally. In very practical terms, I would expect the `handle` function to have the following signatures in these languages:
- **Rust**: `async fn handle(request: OutgoingRequest) -> Result<IncomingResponse, HttpError> {}` (similar to code [here](https://github.com/bytecodealliance/preview2-prototyping/blob/b2382e692bbee7ce189dd23777e76e45abd723b3/host/src/http.rs#L7-L33))
- **TypeScript**: `async function handle(request: OutgoingRequest) -> Promise<Monad<IncomingResponse, HttpError>> {}` (similar to code [here](https://github.com/bytecodealliance/jco/blob/2c5c3bc6bc3ad49f0da9d7ec9e2a44e084432a28/packages/preview2-shim/lib/http/make-request.js))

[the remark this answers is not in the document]
lukewagner: Ok, I'll add a pseudo-future resource (for the pre-`future` Preview2 timeframe) to use in the return type of `outgoing-handler`.  With native `future`/`stream` types, it's easy enough to see how a bindings generator can know to produce `Promise`s and `ReadableStream`s.  But in the Preview2 timeframe, when it's all just opaque resource types (from the binding generator's POV), this adaptation may need to happen manually in handwritten glue code.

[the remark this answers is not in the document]
eduardomourar: Thanks for the explanation. I think I understand better now the difficulties.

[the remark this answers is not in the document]
eduardomourar: If you believe waiting a bit will allow the wit-bindgen tooling to mature the partial async support (at least to a point where a future will not cause an error to be thrown). We can leave this piece to a next iteration of this WIT definition.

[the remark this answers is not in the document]
tyler: Also, we should probably have a couple timeouts built in here. Specifically, an overall timeout and a connect timeout. I believe they should be separate because they will typically have very different sizes, i.e. 1 second vs. 30 seconds, and the correct response to them will differ. For instance, a connect timeout can be safely retried, whereas a normal timeout on a POST request cannot be. 

There are other timeouts we might consider, which each have different uses, such as lookup, time-to-first-byte, time-to-last-byte, and time-between-bytes.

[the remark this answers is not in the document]
lukewagner: Updated with pseudo-future [here](https://github.com/WebAssembly/wasi-http/pull/3/commits/eaa35579fd9c7efb5ca0618eb61ed555548662f8) (and updated wasm-io deps).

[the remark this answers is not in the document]
lukewagner: Added optional timeout parameters [here](https://github.com/WebAssembly/wasi-http/pull/3/commits/2118d730d61ad6e963b9eeb2d173d505a2c1d551).  Initially, I included the timeout options in the `incoming-handler` interface, but because you can't directly chain components pre-Preview3, and thus `incoming-handler` is primarily going to be called by the host, not another component, I couldn't really see how it would be used so I [removed](https://github.com/WebAssembly/wasi-http/pull/3/commits/ec890b2021a6663283b2a83dda3158ce1e421bde) it, but happy to add it back if there is a use.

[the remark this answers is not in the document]
dicej: Just wanted to mention we've had recent requests for concurrent outbound HTTP support in Spin, so we're really excited about your pseudo-future update, @lukewagner!

README.md
```diff
@@ -1,3 +1,38 @@
 # WASI HTTP
 
-(This is a placeholder so there can be a PR to fill in the contents.)
+A proposed [WebAssembly System Interface](https://github.com/WebAssembly/WASI) API.
+
+This proposal currently only contains the proposed Wit interfaces with light
+explanation in comments; more work is necessary to fully document the proposal.
+The Wit comments annotate where the proposed interface is expected to change in
+the short term (for Preview2) once resources and handles are re-added to Wit,
+and then after that (for Preview2) once native stream support is added to the
```
tshepang, line 9: was this meant to be _for Preview3_

lukewagner: Yep!  Thanks, [fixed](https://github.com/WebAssembly/wasi-http/pull/3/commits/36d67b4de2a34536f267bdf66fa8fa48cf9cdb6c)

wit/incoming-handler.wit
```diff
@@ -0,0 +1,24 @@
...
+
+  // The `handle` function takes an outparam instead of returning its response
+  // so that the component may stream its response while streaming any other
+  // request or response bodies. The callee MUST write a response to the
+  // `response-out` and then finish the response before returning. The `handle`
+  // function is allowed to continue execution after finishing the response's
+  // output stream. While this post-response execution is taken off the
+  // critical path, since there is no return value, there is no way to report
+  // its success or failure.
+  handle: func(
```
lann, line 20: We will almost certainly want a way of passing request metadata here, for both "edge" metadata like the remote IP address and to aid component wrapper "middleware" use-cases like authentication.

- Go threads a separate general-purpose "[Context](https://pkg.go.dev/context#WithValue)" argument through call stacks
- Hyper packs metadata into the Request type with "[extensions](https://docs.rs/hyper/0.14.24/hyper/struct.Request.html#method.extensions)".

lukewagner: If we're talking about generic data passed along with a request that isn't part of the HTTP-semantics-defined Fields resource, then I think this might not belong in the HTTP interface (where, in general, this generic data will be sheared off and lost as soon as the request travels over a real HTTP connection) but rather as a (perhaps middleware-specific) interface layered on top of HTTP, defining a new resource type that *contains* an HTTP request (and whatever other contextual metadata).  This also helps make it clear in the component's signature when a component is expecting to send/receive these annotated requests (to/from other components linked into the same middleware framework) vs. when a request is coming/going in/out via standard HTTP semantics.  At least, that's my first impression for how to address this sort of use case.

lann: That seems reasonable for middleware. What do you think about connection metadata like remote IP or peer TLS certs?

brendandburns:
Any chance that we can do a merge and iterate here? Looking through the comments it feels like the major questions (e.g. timeouts) have been resolved, and while there remain minor adjustments, this is the kind of thing that we can adapt as we implement (and indeed watching users actually use a specific implementation may help guide the design)

wit/incoming-handler.wit
lukewagner: Oops, right, I forgot to reply to that part of your last comment.  Potentially these make sense, but lets file separate issues to discuss these independently.  If you would like to do that, that'd be great.

wit/outgoing-handler.wit
```diff
@@ -0,0 +1,18 @@
...
+  use pkg.types.{outgoing-request, request-options, future-incoming-response}
+
+  // The parameter and result types of the `handle` function allow the caller
+  // to concurrently stream the bodies of the outgoing request and the incoming
+  // response.
+  handle: func(
+    request: outgoing-request,
+    options: option<request-options>
+  ) -> future-incoming-response
+}
```
PiotrSikora, line 18: I find those pairings (`incoming-request` and `outgoing-response`, and `outgoing-request` and `incoming-response`) very non-intuitive and a bit redundant.

Could we consider the same prefix on the same connection, e.g. `downstream-request` and `downstream-response`, and `upstream-request` and `upstream-response`?

wit/proxy.wit
```diff
@@ -0,0 +1,45 @@
...
+  //
+  // which will allow a component to import any number of non-default logging
+  // backends that different categories of log messages can be sent to.
+
+  // TODO: add `import metrics: metrics.counters`
+
+  // This is the default handler to use when user code simply wants to make an
+  // HTTP request (e.g., via `fetch()`) but doesn't otherwise specify a
+  // particular handler.
+  import default-upstream-HTTP: pkg.outgoing-handler
```
PiotrSikora, line 29: HTTP world is already overloaded with terms regarding direction, with `upstream`/`downstream` often meaning opposite things in different documents and software. Could we avoid using multiple terms describing the same thing in this spec? Maybe let's stick to `upstream`/`downstream`, so this should be changed to:
`import default-upstream-HTTP: pkg.upstream-handler`

wit/types.wit
```diff
@@ -0,0 +1,158 @@
...
+
+  // This type enumerates the different kinds of errors that may occur when
+  // initially returning a response.
+  variant error {
+      invalid-url(string),
+      timeout-error(string),
+      protocol-error(string),
+      status-error(u16),
+      unexpected-error(string)
+  }
```
PiotrSikora, line 37: This is kind of weird enum, since it mixes status codes with non-HTTP errors. What's the intent here?

Also, the list is pretty limited. Perhaps we should include errors from the `Proxy-Status` RFC (i.e. https://www.iana.org/assignments/http-proxy-status/http-proxy-status.xhtml)?

wit/types.wit
```diff
@@ -0,0 +1,158 @@
...
+
+    // The timeout for the initial connect.
+    connect-timeout-nanos: option<u64>,
+
+    // The timeout for receiving the first byte of the response body.
+    first-byte-timeout-ns: option<u64>,
+
+    // The timeout for receiving the next chunk of bytes in the response body
+    // stream.
+    between-bytes-timeout-ms: option<u32>
```
PiotrSikora, line 111: Those feel very inconsistent with different suffixes, granularity and value types. What's the reason behind that?

wit/types.wit
```diff
@@ -0,0 +1,158 @@
...
+  type incoming-request = u32
+  type outgoing-request = u32
+  drop-incoming-request: func(request: incoming-request)
+  drop-outgoing-request: func(request: outgoing-request)
+  incoming-request-method: func(request: incoming-request) -> method
+  incoming-request-path: func(request: incoming-request) -> string
+  incoming-request-scheme: func(request: incoming-request) -> option<scheme>
+  incoming-request-authority: func(request: incoming-request) -> string
+  incoming-request-headers: func(request: incoming-request) -> headers
+  incoming-request-consume: func(request: incoming-request) -> result<incoming-body>
```
PiotrSikora, line 88: The body is optional, and it's absent in 90%++ of requests.

However, it seems that this API assumes it's always present (presumably, a missing body is represented as an empty stream with EOF, but I don't think that's a good choice), and it requires calls to `consume` and `finish` to get `trailers`, which adds a lot of unnecessary overhead.

wit/outgoing-handler.wit
pchickey: RFC 9110 suggests `inbound` and `outbound` as the preferred terms here, and defines `upstream` and `downstream` differently than you are using them: https://www.rfc-editor.org/rfc/rfc9110#name-intermediaries

PiotrSikora: Right, but the popular implementations (e.g. NGINX and Envoy) disagree with the RFC about `upstream` and `downstream`, and I'm not aware of anything that actually uses the RFC terms.

I considered suggesting `inbound` and `outbound`, but it's yet another term, and it's actually used in service mesh environment to describe traffic direction in a global sense, and not from the perspective of a local endpoint.

pchickey: Our experience at fastly was that `upstream` and `downstream` were so confusing internally, and to our customers, that we stopped using those terms throughout our api and docs.

PiotrSikora: What did you end up replacing it with? I'm not a fan of `downstream` / `upstream` myself (but mostly because the RFCs disagree with implementations), so I'm fine with other terms.

wit/incoming-handler.wit
lann: https://github.com/WebAssembly/wasi-http/issues/4

wit/deps/logging/handler.wit
```diff
@@ -0,0 +1,32 @@
...
+       /// Describes messages likely to be of interest to someone monitoring a
+       /// program.
+       info,
+
+       /// Describes messages indicating hazardous situations.
+       warn,
+
+       /// Describes messages indicating serious errors.
+       error,
+    }
```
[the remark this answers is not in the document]
lukewagner: Yep, that's right, [wasi-logging](https://github.com/WebAssembly/wasi-logging) would be the repo to file that under.

wit/types.wit
```diff
@@ -0,0 +1,158 @@
...
+  fields-clone: func(fields: fields) -> fields
+
+  type headers = fields
+  type trailers = fields
+
+  // The following block defines the `body` type which corresponds to the HTTP
+  // standard Contents. With Preview3, all of these fields can be replaced by a
+  // single type definition:
+  //
+  //   type body = stream<u8, option<trailers>>
```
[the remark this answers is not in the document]
lukewagner: I initially I was going to simply have a method returning a (pseudo) `future<trailers>` (as we discussed a while back), but then I realized that this introduces a tricky question: if I `poll_oneoff` the trailers without concurrently consuming the body stream, does that (1) deadlock (b/c we're not consuming the body, there is backpressure, so we never get the trailers), (2) implicitly discard/ignore the body (so that we can get to the trailers)?  Putting the trailers as the "final value" of the stream avoids this design problem.  But agreed that "`body`" is a misnomer if the type includes trailers.  What do you think of [this](https://github.com/WebAssembly/wasi-http/pull/3/commits/13db7f6e92184d9e4defefc9181043a5d0121b7d) alternative naming that still keeps the same API shape, but avoids calling the composite "body"?

wit/outgoing-handler.wit
lukewagner: Agreed that upstream/downstream is extremely confusing in practice (as evidenced by my own misuse of it below).  I saw incoming/outgoing in one of the other proposals and it seemed more clear when you think of it from the component author's perspective.  That being said, note that all `incoming-`/`outgoing-` prefixes would go away in Preview3 when we get to merge all these resources/interfaces.

wit/proxy.wit
lukewagner: Yep, my mistake mixing and then inverting terminologies.  [Fixed](https://github.com/WebAssembly/wasi-http/pull/3/commits/afec79da5b4d8afec7610dccb77fc4a8cb21b0b8)

wit/types.wit
lukewagner: Good question; I mostly just lifted this from one of the other proposals.  Aligning with HTTP semantically-defined errors sounds generally good, but don't we already get that from the `status-code`?  Perhaps this is something best hashed out in a separate issue.  For now, I've added a [TODO](https://github.com/WebAssembly/wasi-http/pull/3/commits/0bd4751db8c8858f964ecb37dac6cf61bbd82205) to the type definition.

wit/types.wit
lukewagner: Oh gosh, that's embarrassing; it's like I forgot to `git add` before a push or something :P  I aligned them on `ms` [here](https://github.com/WebAssembly/wasi-http/pull/3/commits/297f6536c893f8c316966bc179046523de7280fb), but I think we can discuss further in a follow-up issue if we should change that.

wit/types.wit
lukewagner: Ah, makes sense.  So would a solution be to have `consume` return a `result` containing a 3-case variant such as:
```
variant {
  body(incoming-stream),
  trailers(trailers),
  none
}
```
(and similarly for `write`, but outgoing)?

lukewagner:
Thanks for the great feedback Piotr, you're mostly who I was hoping to hear from before, so I'll merge once we resolve these new comments.

wit/outgoing-handler.wit
brendandburns: Most programming languages have standardized on request/response without incoming/outgoing b/c the context is pretty obvious from whether you are implementing a server or making a client call and programmers don't really get confused about that.

I think we should delete incoming/outgoing or downstream/upstream and just stick to request/response with the context providing the direction.

wit/types.wit
brendandburns: I think we should get rid of `status-error` in general. From the client perspective, if HTTP communication worked, that's not an error, obviously the client surfaces the received status code up to the caller, but it's up to the caller to decide if it is an error.

A classic example of this is whether `404` is an error or not. It really depends on the calling context, and if you force a programmer to do something special when they actually expect a `404` then it becomes problematic.

wit/outgoing-handler.wit
lukewagner: That is definitely the goal with Preview 3.  However, in a Preview 2 timeframe, where we want to do concurrent streaming of requests/responses *and* have blocking `poll_oneoff`, the concrete signatures of the request/response resources have to be different depending on whether you are reading from them or writing to them, so we unfortunately can't wait and infer when the request or response hits a call boundary.  I do believe it's possible to implement source-language libraries which are direction-agnostic (e.g., JS [`Response`](https://developer.mozilla.org/en-US/docs/Web/API/Response)) *in terms of* these these directional resource types, though, so that this is abstracted from the developer (e.g., when constructing a `Response`, the impl would internally call `new-outgoing-response`...).  Avoiding this extra work and complexity is one of the main motivations for adding `future`/`stream` (and the stack-switching required to implement these in the underlying implementation).

wit/types.wit
lukewagner: That makes sense to me.

brendandburns:
@lukewagner so I started to implement this, and I discovered two things:

1) The way it is currently written, you are required to implement the inbound HTTP handler, even if you only want to do outbout HTTP calls.

This code won't compile:

```c
#include "proxy.h"
#include <stdio.h>

int main() {
    default_outgoing_http_outgoing_request_t req;
    default_outgoing_http_request_options_t opts;
    default_outgoing_http_future_incoming_response_t res;

    res = default_outgoing_http_handle(req, &opts);
    printf("FOO: %d\n", res);
}
```

But this code does:

```c
#include "proxy.h"
#include <stdio.h>

void http_handle(uint32_t arg, uint32_t arg0) {

}


int main() {
    default_outgoing_http_outgoing_request_t req;
    default_outgoing_http_request_options_t opts;
    default_outgoing_http_future_incoming_response_t res;

    res = default_outgoing_http_handle(req, &opts);
    printf("FOO: %d\n", res);
}
```

That seems problematic.

Also, it seems that linking in component model code via the `bindgen` macro into `wasmtime` isn't quite working since the existing linking code uses `wasmtime::Linker` and the `bindgen!` macro uses `wasmtime::component::Linker`

Thoughts on how to proceed would be useful/interesting.

Thanks!

Mossaka:
> Also, it seems that linking in component model code via the bindgen macro into wasmtime isn't quite working since the existing linking code uses wasmtime::Linker and the bindgen! macro uses wasmtime::component::Linker

@brendandburns can you try using the `component-model` feature of wasmtime. That will allow you to use `wasmtime::component::Linker`. You can do that by adding this to the "Cargo.toml"

```toml

wasmtime = { workspace = true, features = ["component-model"] }

```

brendandburns:
@Mossaka the existing code paths for the `wasmtime` binary do not make use of the component model linker, and when I tried to instantiate a component model linker (`wasmtime:components:Linker`) pointed at the same `Engine` the existing linker was using my WASM couldn't find the function to link to.

It's possible that there is some name mangling or something going on, but it's hard to debug. I'm happy to share the PR with the implementation if that's useful.

brendandburns:
@lukewagner As I've coded against this more, I'm wondering about whether we really want to store the request and headers options on the host side, rather than just serializing it across as part of the request.

As this is currently written, you cross the guest/host boundary a minimum of 3 times:

1) Allocate a list of headers
2) Allocate the Request object (using the headers from #1)
3) Make the actual request

This doesn't seem very ideal to me as it involves marshalling three different pointers across the boundary. It seems cleaner to just keep everything as strings, and only martial char* across the boundary once.

Thoughts?

danbugs:
> @Mossaka the existing code paths for the `wasmtime` binary do not make use of the component model linker, and when I tried to instantiate a component model linker (`wasmtime:components:Linker`) pointed at the same `Engine` the existing linker was using my WASM couldn't find the function to link to.
> 
> It's possible that there is some name mangling or something going on, but it's hard to debug. I'm happy to share the PR with the implementation if that's useful.

@brendandburns – If you want to look at a couple of example implementations of other proposals (i.e., `wasi-keyvalue`, `wasi-messaging`, and `wasi-sql`), I recommend checking out the following repos:
- https://github.com/Mossaka/keyvalue-component-model-demo ,
- https://github.com/danbugs/wasi-messaging-demo , and
- https://github.com/danbugs/wasi-sql-demo .

~not sure if this helps, but figured I'd drop it here just in case 👍

danbugs (approved):
I think this LGTM after Piotr's comments are addressed~ would be good to merge it and further iterate w/ subsequent PRs 👍

brendandburns:
@danbugs I get that I could build my own binary that uses the component linker, but I don't really want to do that. I want this to be lit up in the main `wasmtime` binary. 

It's not good to have lots of fragmented binaries that each implement different wasi specs, we need to centralize them in a single runtime.

brendandburns:
(and just to be clear, I'm a firm +1 on LGTM/merge + iterate)

eduardomourar (approved):

wit/types.wit
lukewagner: [Fixed](https://github.com/WebAssembly/wasi-http/pull/3/commits/5bac3093efaea7eaacee3f79d6fcf79aad68270d)

lukewagner:
Alright, I'll merge and then file issues for all the remaining issues here and of course everyone else should feel free to do likewise.

wit/types.wit
lukewagner: Filed #5 for the more general question.

wit/types.wit
lukewagner: Filed #6 to consider further.

lukewagner:
And just to follow up on @brendandburns's question above: I think @Mossaka is right that this is related to the issue of optional exports mentioned [here](https://github.com/WebAssembly/component-model/issues/169#issuecomment-1446776193).  The summary is: before too long, we should be able to mark imports and exports as `optional` and, until then, perhaps our tooling should simply treat *all* exports declared in a world as optional (therefore not giving an error if you fail to export one).

That being said, the reason we made `world` a first-class concept in Wit is so that there could be many of them defined, capturing the heterogeneity of embeddings we're all building here.  If a component does not export an HTTP `handle` method, how does it get invoked by the host?  In your example code, the answer is from an exported `main()` function which could correspond to a standard `wasi:cli/main` interface export.  So this would be an entirely different world from `wasi:http/proxy`, e.g.:
```
// app-server.wit
world app-server {
  import default-outgoing-HTTP: pkg.outgoing-handler
  export cli.main
}
```
This world could be included in the wasi-http proposal alongside the `proxy` world and what's useful about having these as two different worlds is it would capture the difference between a host that starts and runs a component in a more traditional app-server-y model vs. a host that is invoking wasm in response to the arrival of HTTP requests in a more serverless model.  (To wit, with the stack suspending/resuming functionality introduced by Preview 3, it should be possible to automatically virtualize both worlds in terms of the other.)

If anyone's interested, I'm happy to create a new issue to dig into this direction more.

brendandburns:
I have begun an implementation here:

https://github.com/bytecodealliance/wasmtime/pull/5929

It is massively incomplete and also hacky, but it (partially) works for some simple examples.

wit/deps/logging/handler.wit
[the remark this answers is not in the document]
sunfishcode: I filed https://github.com/WebAssembly/wasi-logging/issues/9 to track this.

Issue #10, opened 2023-03-07
Query parameters are missing from the WIT


brendandburns:
The definition of outgoing-request defines method, path, authority and scheme but it does not define the query string.

This could be considered part of the path, but typically it is treated separately.

PR #11, opened 2023-03-08
add query field

lukewagner:
It's missing, as pointed out in #10.

This PR brings back some earlier question I had when sketching the initial Wit which I might as well ask since we're in the area:
* Should the request constructor take a single `record` so that the fields are named instead of relying on, in most languages, nameless positional order.
* Should there also be an option to create/consume the {scheme,authority,path,query} fields via single URL?  It does make more URL parsing work for the host (although it seems likely that this code would already be present in the underlying library anyways), but it may factor it out of the guest and may save number-of-hostcalls.  If so, in the request constructor, is the best way to express this as a variant taking either the above record and URL as two cases?

### DIFF
--- a/wit/types.wit
+++ b/wit/types.wit
@@ -79,6 +79,7 @@ default interface types {
   drop-outgoing-request: func(request: outgoing-request)
   incoming-request-method: func(request: incoming-request) -> method
   incoming-request-path: func(request: incoming-request) -> string
+  incoming-request-query: func(request: incoming-request) -> string
   incoming-request-scheme: func(request: incoming-request) -> option<scheme>
   incoming-request-authority: func(request: incoming-request) -> string
   incoming-request-headers: func(request: incoming-request) -> headers
@@ -86,6 +87,7 @@ default interface types {
   new-outgoing-request: func(
     method: method,
     path: string,
+    query: string,
     scheme: option<scheme>,
     authority: string,
     headers: headers


brendandburns:
This PR looks good to me. I am personally in favor of having a single call that takes a URL, at least as an option, this reduces the number of libraries that a web assembly module needs to depend on.

In my opinion, the more dependencies we can push into WASI the better off we are, it makes managing the software supply chain that much easier.

Mossaka (approved):
lgtm!

lukewagner:
Great, I'll merge now and write a new PR for the other two bullets in a bit.

PiotrSikora:
But `query` is not a separate thing at the HTTP layer.

If we're modeling accessors after HTTP/2 pseudo-headers, then `:path` contains both path and query parameters.
From [RFC9113, sec. 8.3.1](https://datatracker.ietf.org/doc/html/rfc9113#section-8.3.1):
> The ":path" pseudo-header field includes the path and query parts of the target URI (the absolute-path production and, optionally, a '?' character followed by the query production; see [Section 4.1](https://www.rfc-editor.org/rfc/rfc9110#section-4.1) of [[HTTP](https://datatracker.ietf.org/doc/html/rfc9110)]).

And if you want to split it, then we need to be very clear that `path` does not include `query` parameters anymore.

Also, what's the return format? A "raw" query-string that needs to be parsed (so kind of pointless split) or a key-value of parsed parameters?

IMHO, this should be reverted.

lukewagner:
Ah interesting.  When I looked at the [URL standard](https://url.spec.whatwg.org/) and saw `pathname` (path) kept separate from `search` (query), I had assumed this would have matched HTTP, but it looks like not.  One thing @tyler and I noticed is that some common implementations will have already parsed and recorded the offset of the query, but not any further structure of the query.  Also, it seems quite common to have path and query split in standard libraries.  But agreed that, if we keep the split, it should be clearly documented that the query is not in the path.  But I'm not really an expert on this topic, so I'm happy to hear other thoughts.

PiotrSikora:
> One thing @tyler and I noticed is that some common implementations will have already parsed and recorded the offset of the query, but not any further structure of the query. Also, it seems quite common to have path and query split in standard libraries.

Right, but hostcalls are not free in Wasm, and we're effectively trading a `path.find('?')` for a hostcall.

lukewagner:
Ah, would your concern be addressed if there was additionally a `path-and-query` getter?  It seems like this is also a common thing to do (e.g. [1](https://docs.rs/hyper/latest/hyper/struct.Uri.html#method.path_and_query), [2](https://learn.microsoft.com/en-us/dotnet/api/system.uri.pathandquery?view=net-6.0)).

PiotrSikora:
That would alleviate my concerns, but I still don't see much benefit in having this split done on the host side.

Also, we should be clear about the scope of `wasi-http`, since this crosses the line between HTTP protocol and creating high-level HTTP APIs and libraries.

Lastly, we also have `wasi-url`, which should be a better home for URI parsing than `wasi-http`.

wit/types.wit
```diff
@@ -79,13 +79,15 @@ default interface types {
   drop-outgoing-request: func(request: outgoing-request)
   incoming-request-method: func(request: incoming-request) -> method
   incoming-request-path: func(request: incoming-request) -> string
+  incoming-request-query: func(request: incoming-request) -> string
```
ereslibre, line 82:
```suggestion
  incoming-request-query: func(request: incoming-request) -> option<string>
```

wit/types.wit
```diff
@@ -79,13 +79,15 @@ default interface types {
...
   incoming-request-scheme: func(request: incoming-request) -> option<scheme>
   incoming-request-authority: func(request: incoming-request) -> string
   incoming-request-headers: func(request: incoming-request) -> headers
   incoming-request-consume: func(request: incoming-request) -> result<incoming-stream>
   new-outgoing-request: func(
     method: method,
     path: string,
+    query: string,
```
ereslibre, line 90:
```suggestion
    query: option<string>,
```

ereslibre (commented):
Sorry for commenting on a merged PR, if the query is to have its own field, it should be optional following the URL Standard wording "[...] any optionally followed by U+003F (?) and a [URL-query string](https://url.spec.whatwg.org/#url-query-string).".

acfoltzer:
@lukewagner the representation `hyper` (and other `http`-crate dependent libraries) uses stores the path and query together in a single bytestring, so reverting this patch likely reduces the overhead for those applications.

lukewagner:
@acfoltzer Probably reverting makes sense, so this is just a technicality but, fwiw, when we poked into it, we found that `PathAndQuery` also stores the offset of the query ([here](https://docs.rs/http/0.2.9/src/http/uri/path.rs.html#14)), so the idea was that this would allow one to slice out just the query when that's all you wanted.  But probably moot point :)

PR #13, opened 2023-03-15
Remove 'query', rename 'path' to 'path-with-query'

lukewagner:
As discussed in #11.  `path-and-query` is made optional as well.  `scheme` was already optional, but `authority` was not, so I made it optional too b/c I think the same rationale applies, but lmk if not.

### DIFF
--- a/wit/types.wit
+++ b/wit/types.wit
@@ -78,18 +78,16 @@ default interface types {
   drop-incoming-request: func(request: incoming-request)
   drop-outgoing-request: func(request: outgoing-request)
   incoming-request-method: func(request: incoming-request) -> method
-  incoming-request-path: func(request: incoming-request) -> string
-  incoming-request-query: func(request: incoming-request) -> string
+  incoming-request-path-and-query: func(request: incoming-request) -> option<string>
   incoming-request-scheme: func(request: incoming-request) -> option<scheme>
-  incoming-request-authority: func(request: incoming-request) -> string
+  incoming-request-authority: func(request: incoming-request) -> option<string>
   incoming-request-headers: func(request: incoming-request) -> headers
   incoming-request-consume: func(request: incoming-request) -> result<incoming-stream>
   new-outgoing-request: func(
     method: method,
-    path: string,
-    query: string,
+    path-and-query: option<string>,
     scheme: option<scheme>,
-    authority: string,
+    authority: option<string>,
     headers: headers
   ) -> outgoing-request
   outgoing-request-write: func(request: outgoing-request) -> result<outgoing-stream>
--- a/wit/types.wit
+++ b/wit/types.wit
@@ -78,14 +78,14 @@ default interface types {
   drop-incoming-request: func(request: incoming-request)
   drop-outgoing-request: func(request: outgoing-request)
   incoming-request-method: func(request: incoming-request) -> method
-  incoming-request-path-and-query: func(request: incoming-request) -> option<string>
+  incoming-request-path-with-query: func(request: incoming-request) -> option<string>
   incoming-request-scheme: func(request: incoming-request) -> option<scheme>
   incoming-request-authority: func(request: incoming-request) -> option<string>
   incoming-request-headers: func(request: incoming-request) -> headers
   incoming-request-consume: func(request: incoming-request) -> result<incoming-stream>
   new-outgoing-request: func(
     method: method,
-    path-and-query: option<string>,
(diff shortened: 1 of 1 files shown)


PiotrSikora (approved):
LGTM, thanks! I have a slight preference for `path-with-query`, but if others prefer `path-and-query` then that's also fine.

brendandburns:
I think that we should just switch everything to URL (and make it non-required) which will resolve all this stuff.

lukewagner:
@PiotrSikora That sounds fine to me, updated.

@brendandburns I think it makes sense to provide URLs as an additional option (both when producing and consuming a request), when that's what you want, but I think there's value in reflecting the underlying protocol fields directly instead of necessarily fusing them into a single URL that, iiuc, doesn't exist at the protocol level.

brendandburns:
@lukewagner If we are interested in helping people by supplying a URL by pieces then combining path and query together are probably the worst API.

While it is perhaps aligned with the spec, this approach is not aligned with how most users consume HTTP based APIs.

Specifically scheme + authority almost never changes, it is very rare to make multiple calls while changing the scheme. However, changing query while keeping path constant _is_ quite common, since the path (e.g. '/api/get-users') will stay constant, while the query string changes (`?userids=foo,bar`) often changes per call depending on input.

lukewagner:
The scheme/authority/path-with-query split is reflected not just in the spec but in implementations as well.  If we have a proper URL to provide all the standard fields (see [this full list](https://developer.mozilla.org/en-US/docs/Web/API/URL)), then I think we'd be entering a no-man's-land by splitting out some, but not all, of the URL fields based on ad hoc judgments of what is common.  Or, at least that seems like a useful design principle to start with before we have more perf or user feedback.

brendandburns:
I guess I would go with prior-art in other programming languages as the right design principle, aka the principle of "least surprise".

In that context the Hyper `Request` builder (https://docs.rs/http/0.2.8/http/request/struct.Builder.html) exclusively takes a URI object. If we look at that URI object, while the `Builder` has exclusively a `path_and_query` constructor, the `Uri` object has getters for `query()`  https://docs.rs/hyper/1.0.0-rc.3/hyper/struct.Uri.html#method.query and `path()`  https://docs.rs/hyper/1.0.0-rc.3/hyper/struct.Uri.html#method.path

Golang URI is similar:
https://pkg.go.dev/net/url#URL.Query
https://pkg.go.dev/net/url#URL.EscapedPath

As is Java:
https://docs.oracle.com/javase/7/docs/api/java/net/URL.html#getPath()
https://docs.oracle.com/javase/7/docs/api/java/net/URL.html#getQuery()

Since this function in the wit is effectively a getter, I think we should follow that pattern.

For the constructor, I'm less certain, but it's worth pointing out that Rust was the only language that I could find where `path_and_query` were explicitly pulled out, most of the constructors just take URL/URIs.

acfoltzer:
> the `Uri` object has getters for `query()` https://docs.rs/hyper/1.0.0-rc.3/hyper/struct.Uri.html#method.query and `path()` https://docs.rs/hyper/1.0.0-rc.3/hyper/struct.Uri.html#method.path

`Uri` also has a getter for [`path_and_query`](https://docs.rs/http/latest/http/uri/struct.Uri.html#method.path_and_query) which returns the underlying representation. I don't think the `http` crate supports this argument very well as the `query` and `path` getters are just views into the primary `PathAndQuery` structure.

Given that both HTTP/2 and HTTP/3 specify the path and query together in the `:path` pesudo-header, I think it's best to stick with the combined `path-and-query` and let the library binding to wasi-http decide whether and how to split them up.

lukewagner:
I agree, and this is why I think `request` should have an accessor that returns a URL and a constructor that takes a URL, because it's not just for paths and queries where we want factor out URL parsing logic, but also all the other subfields of the URL, like individual search params.  And the right place for this seems to be the wasi-url proposal.  But *in addition* to that, I think it's valuable to expose the "ground truth" fields of the HTTP protocol in the HTTP protocol's own terms, which is all this PR is about.

lukewagner:
I should probably wrap this up.  As #11 made this path-query-split change somewhat hastily without the considerations above and since we have agreement from the original proposal champion and other HTTP specialists, I think I'll merge this PR, thereby reverting #11 and improving the naming to avoid the original confusion that led me to misunderstand this the first time around.  Good next steps would be to revive wasi-url and add additional URL constructors/accessors to requests.

PR #15, opened 2023-03-21
Update types.wit

### LINKED ISSUE #14: set-response-outparam needs to take the outparam as a parameter
I'm assuming that we intend for multiple responses to be handled at the same time, if so, the

`set-response-outparam: func(response: result<outgoing-response, error>) -> result`

needs to be

`set-response-outparam: func(outparam: response-outparam, response: result<outgoing-response, error>) -> result`

Because otherwise there is no way to know which outparam the response is tied to.

cc @lukewagner

brendandburns:
Add the outparam to set-outparam-response

Fixes #14

### DIFF
--- a/wit/types.wit
+++ b/wit/types.wit
@@ -118,7 +118,7 @@ default interface types {
   // simply return a `stream`).
   type response-outparam = u32
   drop-response-outparam: func(response: response-outparam)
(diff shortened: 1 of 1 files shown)


lukewagner:
D'oh, thanks!

Issue #18, opened 2023-03-23
Deallocation for responses is unclear.


brendandburns:
It's not entirely clear when in the process a response is actually sent. Is it when `finish-outgoing-stream` is called? Is it when `drop-response` is called (or `drop-outparam`)

We may want to add an explicit `send-response` method.

The reason for this need is because we need to explicitly send the response before we free the resources associated with the stream or the response objects.

cc @lukewagner

lukewagner:
(sorry for slow reply; back now) Good question.  I was expecting that the response starts streaming back as soon as possible, which is when `set-response-outparam` is called.  But this should be explained in the comments, so filed #19.

PR #19, opened 2023-03-27
Explain in more detail how response streaming works

lukewagner:
This attempts to address #18.

### DIFF
--- a/wit/incoming-handler.wit
+++ b/wit/incoming-handler.wit
@@ -12,11 +12,13 @@ default interface incoming-handler {
   // The `handle` function takes an outparam instead of returning its response
   // so that the component may stream its response while streaming any other
   // request or response bodies. The callee MUST write a response to the
-  // `response-out` and then finish the response before returning. The `handle`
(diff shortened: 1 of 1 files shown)


brendandburns:
lgtm.

brendandburns:
fwiw, this is pretty wonky to implement, the reason for this is that the call into the web assembly `handle` routine is synchronous and doesn't return until the complete function executes, which means that effectively you need to spawn the `handle` routine in a different thread and then wait until the `drop` call in order to finalize the HTTP response.

This is feasible, but it's definitely not very clean code.

lukewagner:
Yeah, agreed it's unfortunate, but I think it's mostly unavoidable in the Preview 2 timeframe if we want to support a synchronous/blocking `poll_oneoff`.  In the Preview 3 timeframe, I think we'll be able to do better by moving the fiber-switching (between concurrent `handle` tasks) to be inside the runtime, exposing a normal Rust async interface to the embedder.

Issue #20, opened 2023-03-29
Suggest to use git submodules for deps 


Mossaka:
For the time being, it will run out of sync for deps and their respective repos. I have observed that the `blocking-splice` function is missing from the wasi-io dep in this repo, probably due to the lack of sync. I'd suggest that until we have registry ready for wits, we can use git submodules to keep the deps as subdirectory of this repo.

Mossaka:
I just realized that there isn't really a `wasi-io` repo that contains **only** the wit files, and thus the idea of using `git submodule` will inevitably clone non-wit files down...

lukewagner:
I agree we need to improve our automation here around how we handle deps to get out of this manual-vendoring situation, but yeah, I guess `git submodule` isn't quite an easy fix here.  In the meantime I'll just update the deps manually now and feel free to file PRs that update them in the future (just by copying the files over).

lukewagner:
[Updated](https://github.com/WebAssembly/wasi-http/commit/53e2d258ae3008d254bda79ec958355ded8751a9)

Mossaka:
Per our dicussion from the last component model meeting, I'd like to put this option on the table. Could we tell wit-bindgen that the folders in `deps` are git repos that share the same structures. e.g. in root of each repo, there is a `wit` folder that contains the wit files. This will allow wit-bindgen to fetch the needed wit files and use git submodules to manage these repos. @lukewagner

lukewagner:
That doesn't sound like a bad idea; but I guess this depends on the overall plan we make for the Wit tooling to support these sorts of modular interface development workflows in GitHub.

Mossaka:
Checkout this repo: https://github.com/rvolosatovs/depit

PR #23, opened 2023-04-10
Use `wit-dep` tool to automatically manage dependency vendoring

lukewagner:
This PR uses @rvolosatovs's [`depit`](https://github.com/rvolosatovs/depit) tool to automate the currently-manual steps of copying the Wit documents from other WASI proposal repos.  (Thanks Roman!)  Over time we expect this tooling to evolve further, but this seems like a pretty useful and super-easy-to-adopt short-term improvement.

@Mossaka I think this also goes a long way to addressing the concerns about manual duplication in #20, so I think it resolves that issue too?

(The `world.wit` files are added by actually running `depit` before committing; everything else was in sync already.)

cc @pchickey @sunfishcode @peterhuene

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,9 +20,10 @@ wasm-tools component wit wit/ ...
 
 The `wit/deps` directory contains a live snapshot of the contents of several
 other WASI proposals upon which this proposal depends. It is automatically
-updated by running [`depit`](https://crates.io/crates/depit-cli) in the
-root directory, which fetches the `main` branch of each proposal. As things
-stablize, `wit/deps.toml` will be updated to refer to versioned releases.
+updated by running [`depit update`](https://crates.io/crates/depit-cli) in the
+root directory, which fetches the live contents of the `main` branch of each
+proposal. As things stablize, `wit/deps.toml` will be updated to refer to
+versioned releases.
 
 ### Current Phase
 
--- a/README.md
+++ b/README.md
@@ -18,11 +18,11 @@ or can be manipulated in other ways with:
 wasm-tools component wit wit/ ...
 ```
 
-The HTTP proposal depends on the WASI IO and Logging proposals. For simplicity,
-the Wit files for these proposals are currently copied into the `wit/deps`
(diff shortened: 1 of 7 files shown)


lukewagner:
Ah, good point.  I had wondered whether it was better to use `main` or commit-hashes and chose `main` out of laziness, but maybe I should do commit hashes?

rvolosatovs:
> Ah, good point. I had wondered whether it was better to use `main` or commit-hashes and chose `main` out of laziness, but maybe I should do commit hashes?

Laziness is a virtue, I really love this approach, it requires checking-in `wit/deps` (which is a minuscule price to pay, since that directory only contains WIT definitions), but allows for really simple update procedure and plays well with other tooling.
I've added `depit update` in https://github.com/rvolosatovs/depit/pull/16, which is available in version `0.2.2` https://github.com/rvolosatovs/depit/releases/tag/v0.2.2

lukewagner:
Awesome, thanks!  I just updated the README.md to refer to `depit update`.

README.md
```diff
@@ -18,11 +18,12 @@ or can be manipulated in other ways with:
...
 ```
 
-The HTTP proposal depends on the WASI IO and Logging proposals. For simplicity,
-the Wit files for these proposals are currently copied into the `wit/deps`
-directory and will be updated periodically to match their respective proposals.
-As the Wit tooling develops, we should be able to avoid this form of manual
-vendoring.
+The `wit/deps` directory contains a live snapshot of the contents of several
+other WASI proposals upon which this proposal depends. It is automatically
+updated by running [`depit update`](https://crates.io/crates/depit-cli) in the
```
rvolosatovs, line 23: The project has been moved to https://github.com/bytecodealliance/wit-deps

```suggestion
updated by running [`wit-deps update`](https://crates.io/crates/wit-deps-cli) in the
```

lukewagner: Cool!

Mossaka (approved):
lgtm!

PR #25, opened 2023-04-12
Include wasi-clocks in the proxy world

lukewagner:
This was previously a TODO as I was waiting for the number of clock interfaces to go from 4 to 2, which it now has.


brendandburns:
lgtm.

Issue #24, opened 2023-04-12
WASI-http needs a sequence diagram


brendandburns:
There's a number of functions that allocate/return values (e.g. get a stream for a response body) and methods which dispose of those values (e.g. drop_output_stream).

It's not very clear from the documentation or from the method signatures what is legal and what is a a problem.

e.g. if I call:

```rust
let body_stream = types::incoming_response_consume(incoming_response)
    .map_err(|()| anyhow!("incoming response has no body stream"))?;
types::drop_incoming_response(incoming_response);

let mut body = Vec::new();
let mut eof = false;
while !eof {
    let (mut body_chunk, stream_ended) = streams::read(body_stream, u64::MAX)?;
    eof = stream_ended;
    body.append(&mut body_chunk);
}
streams::drop_input_stream(body_stream);
```

Note I dropped the response before I dropped the stream from the response, is that legal? Who owns the stream? Is it the caller or is it the response object?

Similarly, given the code:
```rust
let request =
    types::new_outgoing_request(method, path, query, Some(scheme), authority, headers);
let request_body = types::outgoing_request_write(request)
    .map_err(|_| anyhow!("outgoing request write failed"))?;
let mut body_cursor = 0;
while body_cursor < body.len() {
    let written =
        streams::write(request_body, &body[body_cursor..]).context("writing request body")?;
    body_cursor += written as usize;
}
streams::drop_output_stream(request_body);

let future_response = default_outgoing_http::handle(request, None);
types::drop_outgoing_request(request);
```

Note that I dropped the output stream before I sent the request, so that seems to imply that the request object owns the output stream.

Given the complexity of the various methods to acquire things like streams and responses and requests and all of the methods to drop them, it is very difficult to know the correct way to either call the code or implement the WIT interface. This means that clients will be confused, and different implementations will behave differently.

We need to either publish a set of acceptance tests that all implementations must be able to run successfully, and/or document the sequence of calls that are "expected" or "acceptable" so that we can have consistency.

lukewagner:
Yeah, we've got a long way to go to properly document this interface and sequence diagrams are a good idea.

Coincidentally, the thing I'm putting finishing touches on now (in the [`child-handles`](https://github.com/WebAssembly/component-model/compare/main...child-handles) branch in the component-model repo) is intended to answer exactly the two questions you're asking above, at the component-model level based solely on the handle type (as opposed to the per-interface level based on documentation).

[...]

PR #26, opened 2023-04-18
Update types.wit with some documentation improvements.

brendandburns:
Ref #24 

cc @lukewagner


eduardomourar (approved):

lukewagner:
Makes sense to me.  I'll wait a bit for more feedback before merging, though.

wit/types.wit
```diff
@@ -73,6 +73,10 @@ default interface types {
   // a single `request` type (that uses the single `stream` type mentioned
   // above). The `consume` and `write` methods may only be called once (and
   // return failure thereafter).
+  // The streams returned by `consume` and `write` are owned by the request and
```
eduardomourar, line 76: Wouldn't the same apply to trailers and headers? Those are also defendant resources that have their lifetime bound to the parent.

brendandburns: I'm less clear that a header is a child resource (or should be a child resource). The reason for this is that the stream of bytes is inherently tied to the TCP connection, but the headers are an in-memory hash-map that has been allocated separately.

In particular, in the case of an outgoing request, the headers are definitively allocated ahead of time and passed into the request by `new-fields` and then passing that object to `new-outgoing-request`

So unless `new-outgoing-request` takes ownership of the fields resource (which is a little wonky imho), the lifetime of the `fields`/`headers` is not tied to a request at all.

lukewagner: That's a good point and I spent some time wrestling with the same question.  I think there are two additional reasons to suggest that, while the allocations are separate, making headers a child resource of the request/response makes sense:
1. From a performance perspective, when a request is passed to `outgoing-handler.handle` or a response is passed to `set-response-outparam`, it would be useful if the host could reliably take ownership of the headers memory so that the host can modify and free the headers as needed without having to keep an original copy in case the guest held onto a reference and accessed the headers in the future. In particular, one scenario that seems likely to happen in a GC language context is that the guest first accesses the headers, creating a GC object wrapper that owns the handle, and then the guest sends off the request/response, but the GC hasn't run yet to finalize the wrapper and drop the handle, so the host is forced to unnecessarily maintain an original copy.
2. From a composability perspective, if I chain two components together as middleware proxies (which seems like it will be a popular application of the http proxy world): are components allowed to hold onto headers when passing on a request/request to the next component?  If yes, then simply mutating headers for a request that you currently own may possibly break other components in unpredictable ways, so it seems like the answer should be no (or else noone can mutate headers without conservatively making a clone first), which returning a child handle ensures.

wit/types.wit
```diff
@@ -95,6 +95,10 @@ interface types {
   // are shared with the request, with all mutations visible to all uses.
   // Components MUST avoid updating `headers` and `trailers` after passing a
   // request that points to them to the outside world.
+  // The streams returned by `consume` and `write` are owned by the request and
+  // response objects. The streams are destroyed when the request/response is
+  // dropped. The caller can also call drop on the stream before the
+  // request/response is dropped if they want to release resources earlier.
```
lukewagner, line 101:
```suggestion
  // dropped, thus a client MUST drop any handle referring to a request/response stream
  // before dropping the request/response or passing ownership of the request/response
  // to the outside world. The caller can also call drop on the stream before the 
  // request/response is dropped if they want to release resources earlier.
```

lukewagner: Sounds good, one last suggested tweak.

brendandburns:
@lukewagner sorry for the delay, added your suggestion. Thanks!

lukewagner:
Thanks!

PR #29, opened 2023-04-19
Change fields value type from string to list<u8>

### LINKED ISSUE #28: Field values aren't guaranteed to be ASCII
The HTTP spec _recommends_ but does not _require_ that field values are ASCII:

https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html#rfc.section.5.5.p.4

> Field values are usually constrained to the range of US-ASCII characters [USASCII]. Fields needing a greater range of characters can use an encoding, such as the one defined in [RFC8187]. Historically, HTTP allowed field content with text in the ISO-8859-1 charset [ISO-8859-1], supporting other charsets only through use of [RFC2047] encoding. Specifications for newly defined fields SHOULD limit their values to visible US-ASCII octets (VCHAR), SP, and HTAB. A recipient SHOULD treat other allowed octets in field content (i.e., obs-text) as opaque data.

Currently, field values are represented by `string`s, but this leaves some ambiguity in how to represent non-ASCII values (e.g. `"\xff"`). A couple of options:

- Change the field value type to `list<u8>`
- Document that field values are transcoded from latin1

lann:
Closes #28

### DIFF
--- a/wit/types.wit
+++ b/wit/types.wit
@@ -43,11 +43,11 @@ default interface types {
   type fields = u32
   drop-fields: func(fields: fields)
   new-fields: func(entries: list<tuple<string,string>>) -> fields
-  fields-get: func(fields: fields, name: string) -> list<string>
-  fields-set: func(fields: fields, name: string, value: list<string>)
+  fields-get: func(fields: fields, name: string) -> list<list<u8>>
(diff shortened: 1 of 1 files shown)


pchickey (approved):

lukewagner:
Sounds good to me.  I'll wait a bit for comment before merging.

PR #31, opened 2023-05-10
switch from wasi-logging to wasi-cli for console output

lukewagner:
This PR switches the proxy world to use the stdout/stderr interfaces from wasi-cli for the default console output.  This way all the normal wasi-libc machinery that sends libc writes to stdout/stderr will Just Work without requiring manual user work or per-world libc configuration.


pchickey:
The only concern I have with this is that `stdio.stdin` is not provided by the world. Clearly, it doesn't make sense to have stdin actually provide input to this sort of application, but I am not certain that we can get a libc (or preview 1 adapter) which will not need that import available. This might mean the user will have to do some sort of work to virtualize a trivial null stdin device (the stream which always returns EOF) onto their component before it will run in this world.

While I hope that in the long run we end up with the sort of tooling where that is trivial, I don't want it to be a stumbling block as we start out.

Could we have the proxy world provide that import, noting in a comment that it should always be an empty stream, is just provided for libc compatibility, and is expected to be removed in the preview 3 timeframe?

lukewagner:
I worry that adding `stdio.stdin` will entice hosts to think of creative and non-interoperable things to do with it other than simply returning eof.  Is there not something that wasi-libc can do with weak symbols or that wit-component can do with recognizing and stubbing out?  Maybe @sunfishcode has some ideas?  Because this seems like a pretty unfortunate low-level implementation detail to bubble up all the way to the world.

sunfishcode:
Yes, there are weak-symbol ways we can make wasi-libc avoid depending on stdin imports when stdin isn't used. But as @pchickey says, that is more work to implement, and seems like something we could defer to preview3.

pchickey:
Even with weak symbols, libc needs to support when e.g. `read(2)` is called with `STDIN_FILENO`, so effectively all programs using any file reading functionality in libc will end up needing the stdin symbols at link time, even if they are never actually used at runtime.

lukewagner:
Talking through more options with @sunfishcode, we had another idea for how to make this painless for wasi-libc: wasi-libc unconditionally imports the `get-stdin` function and that propagates all the way into the final core `.wasm`, and then if wit-component sees this import and sees that `stdin` is not present in the target world, it just mocks up the stub implementation that returns the eof stdin stream.  What's neat about this approach is that I think it's also what we'd want to do for uses of wasi-filesystem from wasi-libc when the target world doesn't support wasi-filesystem (as is the case in wasi-http and wasi-cloud-core). Could that work?

pchickey:
That seems like a reasonable fix, but I won't have time to implement that functionality for a month or two. Can we start by keeping stdin in, and file an issue to work on the wit-componentize tooling and eliminate it once that work is done?

lukewagner:
Righto, done.

sunfishcode:
Do you know why this PR is adding a copy of wasi-sockets to the deps directory? wasi-http isn't expected to include sockets, so it's surprising that there would be a wit/deps/sockets directory with the sockets files in it.

lukewagner:
Yeah, it's unfortunate, but when I use `wasm-tools component wit wit/ -t` to check the validity of the Wit, it parses `command.wit`  and issues a validation error if I don't have `sockets`, even though `command.wit` and `sockets` don't actually show up anywhere in the output of `wasm-tools component wit wit/ -t`.  (I have a little comment explaining this in `wit/deps.toml`.)

pchickey (approved):
Sorry, missed that this was waiting on an explicit +1

PR #32, opened 2023-05-18
Update to the latest wasi-random.

sunfishcode:
The main change here is to add the new "insecure" API, and also to split random, insecure, and insecure-seed into separate interfaces.


Mossaka:
@sunfishcode you need to rebase to the new wit package syntax.

sunfishcode:
The updates to wasi-random were included in the overall updates, so this PR is no longer needed.

PR #35, opened 2023-05-30
Add CI check to veirfy deps are up-to-date

Mossaka:
This PR adds a GH Action to check if deps are up-to-date. It can also verify all the wit-abis if they are updated to use the newest wit package syntax.


pchickey (approved):
Thanks!

PR #45, opened 2023-06-20
Make new request/response functions fallible

### LINKED ISSUE #44: new-outgoing-request should return a Result
There are lots of reasons why a call to `new-outgoing-request` could fail, right now there's no way to tell the guest why something failed other than returning a NULL pointer.

lukewagner:
Resolves #44

### DIFF
--- a/wit/types.wit
+++ b/wit/types.wit
@@ -111,7 +111,7 @@ interface types {
     scheme: option<scheme>,
     authority: option<string>,
     headers: headers
-  ) -> outgoing-request
+  ) -> result<outgoing-request, error>
   outgoing-request-write: func(request: outgoing-request) -> result<outgoing-stream>
 
   // Additional optional parameters that can be set when making a request.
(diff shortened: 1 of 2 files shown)


Mossaka (approved):
lgtm!

wit/types.wit
```diff
@@ -111,7 +111,7 @@ interface types {
     scheme: option<scheme>,
     authority: option<string>,
     headers: headers
-  ) -> outgoing-request
+  ) -> result<outgoing-request, error>
```
Mossaka, line 114: nit: consider adding a line of comment describing the failure cases.

lukewagner: You're right, but, realistically, someone with a better, deeper understanding of HTTP probably needs to write this comment and describe what precisely is checked eagerly.  More generally, the current comments are not spec-level in precision and formality--they're mostly just enough to let an implementor see how things are supposed to hang together--so we'll need a full pass over the whole interface to get this to a spec-level of quality at some point in the future.

brendandburns: I think it is sufficient to just return a result. The error that is returned will contain the provider specific issues. Some providers (e.g. embedded) may not load the certificate roots to support https for example, while other clients will support that.

LGTM.

PR #47, opened 2023-08-03
fix: update wit dependencies to latest

eduardomourar:
This PR will just update to the latest WIT definitions from other WASI proposals.


lukewagner:
Thanks!  Looks like the wit ABI doc has to be updated (see message in the failed CI check above), and then I'll merge.

eduardomourar:
Done! (I was running using a newer version of wit-bindgen).

PR #48, opened 2023-08-22
fix: rename `types` to `http-types`

rvolosatovs:
This is done for compatiblity with Wasmtime 11 and the rest of Wasm tooling


rvolosatovs:
whoops, wrong base repo, sorry

PR #50, opened 2023-08-28
refactor: use include syntax for wasi:clocks

Mossaka:
This commit uses `include` syntax to simplify the proxy world by hiding the interfaces from wasi:clock package


lukewagner:
There's an interesting usability question here for how aggressively we want to use `include` to "factor out" imports between different proposals.  For example, given the large number of interfaces in the WASI sockets proposal, `include` seems like a clear win.  OTOH, if a proposal had one interface and one world, it would only add indirection (for the reader) to `include` the world instead of `import`ing the interface.  Cases like this PR are somewhere in the middle. 

My impression is that, just looking at the before/after diff, the before state is more readable.  However, I think that might just be superficially due to the name of the world: `include wasi:clocks/example-world` doesn't really tell me much.  Just to pick a maybe-too-cute example, `include wasi:clocks/support` feels like it would be net more readable, so maybe we could pick that or some more helpful name than `wasi:clocks/example-world`?

Mossaka:
Yeah totally agree that `example-world` isn't helpful. Pat made a suggestion to rename all the `example-world` to `imports` and I have a PR for each one of them (e.g. wasi:clocks: https://github.com/WebAssembly/wasi-clocks/pull/49). If we think `support` world would be a better name, I am more than happy to change it.

pchickey:
in a separate PR I suggested we rename `example-world` to `imports`. this is appropriate for poll, io, filesystem, sockets, clocks, and random. we should not export an `imports` world for http and cli.

Mossaka:
I've updated the PR to use `wasi:clocks/imports` world.

wit/deps/sockets/ip-name-lookup.wit
```diff
@@ -1,6 +1,6 @@
 
 interface ip-name-lookup {
-	use wasi:poll/poll.{pollable}
+	use wasi:io/poll.{pollable}
```
pchickey, line 3: I think somewhere the upgraded deps are out of sync - this one already has the merge of `poll` into the `wasi:io` package, but that isnt reflected in the rest of this tree.

Mossaka:
Rebased

wit/deps/sockets/ip-name-lookup.wit
Mossaka: Should be fixed by now.

wit/proxy.wit
```diff
@@ -5,10 +5,8 @@ package wasi:http;
 /// this world may concurrently stream in and out any number of incoming and
 /// outgoing HTTP requests.
 world proxy {
-  /// HTTP proxies have access to time and randomness.
-  import wasi:clocks/wall-clock;
-  import wasi:clocks/monotonic-clock;
-  import wasi:clocks/timezone;
+  // HTTP proxies have access to time and randomness.
```
lukewagner, line 8:
```suggestion
  /// HTTP proxies have access to time and randomness.
```
/me shakes fist at >1 kinds of comments

lukewagner (approved):
lgtm

pchickey (approved):

Issue #49, opened 2023-08-22
consider defining `error` in terms of RFC 9209


lukewagner:
There is a [TODO](https://github.com/WebAssembly/wasi-http/blob/main/wit/types.wit#L29-L37) to properly fill in the `error` enum since there's nothing we can simply point to in the [HTTP Semantics spec](https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html) like we do for pretty much the rest of `wasi:http/types`.  At the moment it's just a totally ad hoc collection of cases, which doesn't feel great.  @acfoltzer pointed out that the [Error Types section of RFC 9209](https://www.rfc-editor.org/rfc/rfc9209#name-proxy-error-types) could be a useful implementation-neutral standards basis for defining the error cases, so I'd like to consider doing that or hear if anyone else has any better ideas.

PiotrSikora:
I'm glad that RFC 9209 is being used, but you should probably refer to the [IANA registry for HTTP Proxy Error Types](https://www.iana.org/assignments/http-proxy-status/http-proxy-status.xhtml) directly to account for future error types.

Also, keep in mind that it's not a "complete" solution, because `http_request_error` piggy-backs on the existing HTTP status codes instead of redefining them as standalone error types (unfortunately, I lost that battle at IETF).

acfoltzer:
> Also, keep in mind that it's not a "complete" solution, because `http_request_error` piggy-backs on the existing HTTP status codes instead of redefining them as standalone error types (unfortunately, I lost that battle at IETF).

@PiotrSikora wouldn't failures of that type be reflected in `wasi-http` as an "ok" result with the appropriate 4xx/5xx status? Are there applications where it would be useful to have status codes in `error`?

PiotrSikora:
> @PiotrSikora wouldn't failures of that type be reflected in `wasi-http` as an "ok" result with the appropriate 4xx/5xx status?

No, local errors shouldn't be mimicked as fake responses with status codes, otherwise you cannot distinguish WASI errors from real errors received from the remote.

> Are there applications where it would be useful to have status codes in `error`?

It's not about having status codes in errors. My point was that there is a whole set of errors missing from that registry, because the HTTP working group didn't want to duplicate those that already had a defined HTTP status code.

e.g. when dispatching a new request with payload above the host's configured size limits, host might want to reject the call with "payload too large" or similar error, but that's not available in the HTTP Proxy Error Types because `413 Payload Too Large` already exists.

[...]

PR #52, opened 2023-09-12
Fill in the 'error' variant

lukewagner:
As discussed in #49, this PR replaces the current placeholder `error` variant cases with the error types enumerated by [RFC 9209](https://www.rfc-editor.org/rfc/rfc9209#name-proxy-error-types) which are maintained in an [IANA table](https://www.iana.org/assignments/http-proxy-status/http-proxy-status.xhtml#table-http-proxy-error-types).  Note that these specs also specify payloads for some of these error cases which I also reflected in the `error` case payloads.

The one interesting bit of minutiae, where I'm going with [Adam's suggestion](https://github.com/WebAssembly/wasi-http/issues/49#issuecomment-1693858871), but I have no strong opinions about so happy to hear arguments to the contrary, is that there is no `error` case corresponding to `http_request_error`, as this seems like it would turn into an `ok` `result` with a 4xx `status-code`.

### DIFF
--- a/wit/types.wit
+++ b/wit/types.wit
@@ -27,14 +27,65 @@ interface types {
     other(string)
   }
 
-  /// TODO: perhaps better align with HTTP semantics?
-  /// This type enumerates the different kinds of errors that may occur when
-  /// initially returning a response.
+  // The cases of this variant correspond to the IANA HTTP Proxy Error Types:
+  //   https://www.iana.org/assignments/http-proxy-status/http-proxy-status.xhtml#table-http-proxy-error-types
   variant error {
-    invalid-url(string),
-    timeout-error(string),
-    protocol-error(string),
-    unexpected-error(string)
+    DNS-timeout,
+    DNS-error(DNS-error-payload),
+    destination-not-found,
+    destination-unavailable,
+    destination-IP-prohibited,
+    destination-IP-unroutable,
+    connection-refused,
+    connection-terminated,
+    connection-timeout,
+    connection-read-timeout,
+    connection-write-timeout,
+    connection-limit-reached,
+    TLS-protocol-error,
+    TLS-certificate-error,
+    TLS-alert-received(TLS-alert-received-payload),
+    HTTP-request-error(HTTP-request-error-payload),
+    HTTP-request-denied,
+    HTTP-response-incomplete,
+    HTTP-response-header-section-size(u32),
+    HTTP-response-header-size(field-size-payload),
+    HTTP-response-body-size(u32),
+    HTTP-response-trailer-section-size(u32),
+    HTTP-response-trailer-size(field-size-payload),
+    HTTP-response-transfer-coding(string),
+    HTTP-response-content-coding(string),
+    HTTP-response-timeout,
+    HTTP-upgrade-failed,
+    HTTP-protocol-error,
+    proxy-internal-response,
+    proxy-internal-error,
+    proxy-configuration-error,
+    proxy-loop-detected
+  }
+
+  // Defines the case payload type for `DNS-error` above:
+  record DNS-error-payload {
+    rcode: string,
+    info-code: u16
+  }
+
+  // Defines the case payload type for `TLS-alert-received` above:
+  record TLS-alert-received-payload {
+    alert-ID: u8,
+    alert-message: string
+  }
+
+  // Defines the case payload type for `HTTP-request-error` above:
+  record HTTP-request-error-payload {
+    status-code: u16,
+    status-phrase: string
+  }
+
+  // Defines the case payload type for `HTTP-response-{header,trailer}-size` above:
+  record field-size-payload {
+    field-name: string,
+    field-size: u32
   }
 
   /// This type enumerates the different kinds of errors that may occur when
--- a/wit/types.wit
+++ b/wit/types.wit
@@ -45,8 +45,12 @@ interface types {
     TLS-protocol-error,
     TLS-certificate-error,
     TLS-alert-received(TLS-alert-received-payload),
-    HTTP-request-error(HTTP-request-error-payload),
     HTTP-request-denied,
+    HTTP-request-length-required,
+    HTTP-request-content-too-large,
+    HTTP-request-URI-too-long,
+    HTTP-request-header-section-size(option<u32>),
+    HTTP-request-header-size(option<field-size-payload>),
     HTTP-response-incomplete,
     HTTP-response-header-section-size(option<u32>),
     HTTP-response-header-size(field-size-payload),
@@ -74,12 +78,6 @@ interface types {
   record TLS-alert-received-payload {
     alert-ID: option<u8>,
     alert-message: option<string>
-  }
-
-  /// Defines the case payload type for `HTTP-request-error` above:
-  record HTTP-request-error-payload {
-    status-code: option<u16>,
-    status-phrase: option<string>
   }
 
   /// Defines the case payload type for `HTTP-response-{header,trailer}-size` above:
--- a/wit/types.wit
+++ b/wit/types.wit
@@ -27,7 +27,7 @@ interface types {
     other(string)
   }
 
-  /// The cases of this variant correspond to the IANA HTTP Proxy Error Types:
+  /// These cases are inspired by the IANA HTTP Proxy Error Types:
   ///   https://www.iana.org/assignments/http-proxy-status/http-proxy-status.xhtml#table-http-proxy-error-types
   variant error-code {
     DNS-timeout,
@@ -47,14 +47,18 @@ interface types {
     TLS-alert-received(TLS-alert-received-payload),
     HTTP-request-denied,
     HTTP-request-length-required,
-    HTTP-request-content-too-large,
+    HTTP-request-body-size(option<u64>),
+    HTTP-request-method-invalid,
+    HTTP-request-URI-invalid,
     HTTP-request-URI-too-long,
     HTTP-request-header-section-size(option<u32>),
     HTTP-request-header-size(option<field-size-payload>),
+    HTTP-request-trailer-section-size(option<u32>),
+    HTTP-request-trailer-size(field-size-payload),
     HTTP-response-incomplete,
     HTTP-response-header-section-size(option<u32>),
     HTTP-response-header-size(field-size-payload),
-    HTTP-response-body-size(option<u32>),
+    HTTP-response-body-size(option<u64>),
     HTTP-response-trailer-section-size(option<u32>),
     HTTP-response-trailer-size(field-size-payload),
     HTTP-response-transfer-coding(option<string>),
@@ -62,10 +66,14 @@ interface types {
     HTTP-response-timeout,
     HTTP-upgrade-failed,
     HTTP-protocol-error,
-    proxy-internal-response,
-    proxy-internal-error,
-    proxy-configuration-error,
-    proxy-loop-detected
+    loop-detected,
+    configuration-error,
+    /// This is a catch-all error for anything that doesn't fit cleanly into a
+    /// more specific case. It also includes an optional string for an
+    /// unstructured description of the error. Users should not depend on the
+    /// string for diagnosing errors, as it's not required to be consistent
+    /// between implementations.
+    internal-error(option<string>)
   }
 
   /// Defines the case payload type for `DNS-error` above:
@@ -76,7 +84,7 @@ interface types {
 
   /// Defines the case payload type for `TLS-alert-received` above:
   record TLS-alert-received-payload {
-    alert-ID: option<u8>,
+    alert-id: option<u8>,
     alert-message: option<string>
   }
 
--- a/wit/types.wit
+++ b/wit/types.wit
@@ -27,8 +27,8 @@ interface types {
     other(string)
   }
 
-  // The cases of this variant correspond to the IANA HTTP Proxy Error Types:
-  //   https://www.iana.org/assignments/http-proxy-status/http-proxy-status.xhtml#table-http-proxy-error-types
+  /// The cases of this variant correspond to the IANA HTTP Proxy Error Types:
+  ///   https://www.iana.org/assignments/http-proxy-status/http-proxy-status.xhtml#table-http-proxy-error-types
   variant error {
     DNS-timeout,
     DNS-error(DNS-error-payload),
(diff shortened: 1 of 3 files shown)


Mossaka (approved):
lgtm!

wit/types.wit
```diff
@@ -26,14 +26,60 @@ interface types {
     other(string)
   }
 
-  // TODO: perhaps better align with HTTP semantics?
-  // This type enumerates the different kinds of errors that may occur when
-  // initially returning a response.
+  // The cases of this variant correspond to the IANA HTTP Proxy Error Types:
+  //   https://www.iana.org/assignments/http-proxy-status/http-proxy-status.xhtml#table-http-proxy-error-types
+  // Note: the `http_request_error` case maps to an `ok` response with a 4xx
+  // `status-code`.
```
PiotrSikora, line 32: How exactly would that happen? Do we expect hosts to generate fake responses instead of returning `error`?

lukewagner: (Depending on how recent discussion in #49 resolves, it sounds like this comment might just go away and we'll add an `http-request-error` instead.)

lukewagner: ... and that's what I did in [this](https://github.com/WebAssembly/wasi-http/pull/52/commits/d25c0ab39b9b1818a414337c0ce8a18c3e7b5553) last commit.  @acfoltzer make sense to you too?

PiotrSikora: I don't think returning e.g. `http-request-error(413)` makes for a great developer experience when host rejects outgoing HTTP request because of size limits.

lukewagner: What do you think should happen in that case?

pchickey:
This PR should also add an `http-error: func(stream-error) -> option<error>` function for elaborating `stream.error` into one of these HTTP-specific error variants.

wit/types.wit
```diff
@@ -26,14 +26,65 @@ interface types {
...
+  record HTTP-request-error-payload {
+    status-code: u16,
+    status-phrase: string
+  }
+
+  // Defines the case payload type for `HTTP-response-{header,trailer}-size` above:
+  record field-size-payload {
+    field-name: string,
+    field-size: u32
   }
```
acfoltzer, line 88: The fields corresponding to extra parameters in RFC 9209 should all be optional

wit/types.wit
acfoltzer: I think we need to add more variants to cover the aspects of an `outgoing-request` that the handler implementation might want or have to reject. The `proxy-status` RFC is not concerned with things like request URI size limits, because the proxy can just return a 414 directly to the client, so we shouldn't expect that it'll comprehensively cover the cases we're concerned with.

If we sufficiently cover the reasons that an embedder might reject an `outgoing-request` through variants in this error, enforcement when constructing the `outgoing-request`, setting the request's headers, or writing to its body, then I don't think it makes sense to include `http-request-error` for the reasons @PiotrSikora mentions.

@lukewagner, we've recently discussed a slice of this in our discussions of header filtering—this is one reason the embedder might want to reject a request. We should consider adding direct variants that would correspond to use of `HTTP-request-error` with 4xxs that speak to the other potential reasons an embedder may need to reject:

- `HTTP-request-length-required` (corresponds to 411)
- `HTTP-request-content-too-large` (corresponds to 413)
- `HTTP-request-URI-too-long` (corresponds to 414)
- `HTTP-response-header-section-size(u32)` and `HTTP-response-header-size(field-size-payload)` and trailer variants (corresponds to 431)

Some of the other 4xx codes that `proxy-status` contemplates for `HTTP-request-error` include codes like 403 that I think deserve a more specific reason if the embedder would be tempted to use it (like the presence of an illegal header), or codes like 405 which speak to properties that only the origin can know.

wit/types.wit
pchickey: to clarify: is each row in these records optional, or is the presence of the record itself in the variant case optional?

wit/types.wit
```diff
@@ -26,14 +26,65 @@ interface types {
...
+    HTTP-response-header-section-size(u32),
+    HTTP-response-header-size(field-size-payload),
+    HTTP-response-body-size(u32),
+    HTTP-response-trailer-section-size(u32),
+    HTTP-response-trailer-size(field-size-payload),
+    HTTP-response-transfer-coding(string),
+    HTTP-response-content-coding(string),
+    HTTP-response-timeout,
+    HTTP-upgrade-failed,
+    HTTP-protocol-error,
```
pchickey, line 59: ` This error should only be used when a more specific one is not defined.` from https://www.iana.org/assignments/http-proxy-status/http-proxy-status.xhtml#table-http-proxy-error-types - this can represent whatwg fetch's `network error` concept https://fetch.spec.whatwg.org/#concept-network-error. Maybe this should have an `(option<string>)` payload so that an implementation can provide extra debugging information if it is available? and the docs should make it clear this is the catch-all case for when no other error could be determined.

wit/types.wit
acfoltzer: @pchickey each row is optional; there is no record in the concrete RFC 9209 definition

wit/types.wit
acfoltzer: > Maybe this should have an `(option<string>)` payload so that an implementation can provide extra debugging information if it is available?

I worry that if we expose this extra debugging information to clients, this will become load-bearing like the situation we have with 503s and custom HTTP/1 reason phrases. Perhaps we could instead encourage embedders to treat this as a cause for internal logging.

lukewagner:
This all looks great to me; thanks for all the work @elliottt!  (I'd approve, but I can't as the PR author :P)

acfoltzer:
@lukewagner can we please make sure this is addressed before committing? https://github.com/WebAssembly/wasi-http/pull/52#discussion_r1379277447

wit/types.wit
```diff
@@ -27,16 +27,77 @@ interface types {
...
+    connection-limit-reached,
+    TLS-protocol-error,
+    TLS-certificate-error,
+    TLS-alert-received(TLS-alert-received-payload),
+    HTTP-request-denied,
+    HTTP-request-length-required,
+    HTTP-request-content-too-large,
+    HTTP-request-URI-too-long,
+    HTTP-response-header-section-size(option<u32>),
+    HTTP-response-header-size(option<field-size-payload>),
```
acfoltzer, line 53: @elliottt 
```suggestion
    HTTP-request-header-section-size(option<u32>),
    HTTP-request-header-size(option<field-size-payload>),
```

elliottt: Thanks!

wit/types.wit
pchickey: No, I think we really need to provide some option<string> here or else we will regret it. For the catch-all error case, we need to have some sort of way for the user to debug why an unexpected error occurred. Like the stream error `to-debug-string`, we will document that this information is unstable. If programmers ignore that and it breaks their code when it changes, I would prefer that situation to programmers whose code breaks when they move it to a different embedding, but they have no reasonable means to determine why, especially given that the purpose of this standard is to work on many different embedding which will have a variety of constraints which we can't yet imagine. Sending a user off to dig an error out of internal logging is always a much worse solution than just providing a string.

wit/types.wit
```diff
@@ -27,16 +27,77 @@ interface types {
...
+    connection-timeout,
+    connection-read-timeout,
+    connection-write-timeout,
+    connection-limit-reached,
+    TLS-protocol-error,
+    TLS-certificate-error,
+    TLS-alert-received(TLS-alert-received-payload),
+    HTTP-request-denied,
+    HTTP-request-length-required,
+    HTTP-request-content-too-large,
```
PiotrSikora, line 50: The corresponding error for the response is `HTTP-response-body-size(option<u32>)`, so this introduces inconsistency within the spec.

wit/types.wit
```diff
@@ -27,16 +27,77 @@ interface types {
...
+    HTTP-request-URI-too-long,
+    HTTP-request-header-section-size(option<u32>),
+    HTTP-request-header-size(option<field-size-payload>),
```
PiotrSikora, line 53: We should also add:
```
HTTP-request-trailer-section-size(option<u32>),
HTTP-request-trailer-size(field-size-payload),
```

wit/types.wit
```diff
@@ -27,16 +27,77 @@ interface types {
...
+    connection-terminated,
+    connection-timeout,
+    connection-read-timeout,
+    connection-write-timeout,
+    connection-limit-reached,
+    TLS-protocol-error,
+    TLS-certificate-error,
+    TLS-alert-received(TLS-alert-received-payload),
+    HTTP-request-denied,
+    HTTP-request-length-required,
```
PiotrSikora, line 49: The generic `HTTP-request-error` would be still useful, since those listed here don't even cover the existing 4xx error space. Either that, or we should add `HTTP-request-*` errors covering existing 4xx error codes.

wit/types.wit
```diff
@@ -27,16 +27,77 @@ interface types {
...
+    HTTP-response-header-size(field-size-payload),
+    HTTP-response-body-size(option<u32>),
+    HTTP-response-trailer-section-size(option<u32>),
+    HTTP-response-trailer-size(field-size-payload),
+    HTTP-response-transfer-coding(option<string>),
+    HTTP-response-content-coding(option<string>),
+    HTTP-response-timeout,
+    HTTP-upgrade-failed,
+    HTTP-protocol-error,
+    proxy-internal-response,
```
PiotrSikora, line 65: Note that this is used to indicate that the intermediate generated the response and didn't forward the request to the next hop. It's not an error condition.

wit/types.wit
```diff
@@ -27,16 +27,77 @@ interface types {
...
+    proxy-internal-error,
+    proxy-configuration-error,
```
PiotrSikora, line 67: I'm not sure if those are applicable here?

wit/types.wit
```diff
@@ -27,16 +27,77 @@ interface types {
...
-  /// TODO: perhaps better align with HTTP semantics?
-  /// This type enumerates the different kinds of errors that may occur when
-  /// initially returning a response.
-  variant error {
-    invalid-url(string),
-    timeout-error(string),
-    protocol-error(string),
-    unexpected-error(string)
+  /// The cases of this variant correspond to the IANA HTTP Proxy Error Types:
+  ///   https://www.iana.org/assignments/http-proxy-status/http-proxy-status.xhtml#table-http-proxy-error-types
```
PiotrSikora, line 31: The list below already diverged from the IANA table.

wit/types.wit
acfoltzer: @pchickey I think it comes down to who the "user" is that we want to empower to debug. If the user is a person who moves their code from one hosting provider to another and suddenly they start seeing protocol errors for requests to the same origin, that is very unlikely to be actionable by that user. It almost certainly indicates a problem with the embedder or the network between the embedder and the origin. In this situation, adding a free-form string at best gives that user something to paste into the support request with their new provider.

If you have other user profiles in mind that would benefit from the debugging info directly, I don't want to hold up progress on the spec. I would just recommend that any embedder who wants to avoid ossification _not_ populate this field for production traffic.

wit/types.wit
acfoltzer: My rationale [laid out here](https://github.com/WebAssembly/wasi-http/pull/52#discussion_r1379277447) is that the remaining 4xx space would best be covered by returning an `ok` result with the next hop's 4xx response directly conveyed. Unlike the consumer of a `proxy-status` header, the perspective of the `wasi-http` client code makes it clear via the `result` whether the error arises from within the embedder, or from the next hop, so I don't think we need this case which primarily exists to disambiguate whether it was a proxy-generated status code or the next hop's.

Following the addition of the `HTTP-request-*` variants, are there further 4xx codes that you see being ambiguous? That's probably an indication that we're missing variants still.

wit/types.wit
acfoltzer: Yeah, I think this variant probably doesn't belong, especially because returning an error means we _don't_ get a response for this tag to apply to.
```suggestion
```

wit/types.wit
acfoltzer: I think it's useful to have a catch-all "internal error" when implementing an embedder, and setting apart whether that error might have arisen from a bad configuration is helpful too. But we should probably not label them as being specific to a `proxy`; they could equally apply to incoming- or outgoing-only handler programs.
```suggestion
    internal-error,
    configuration-error,
```

wit/types.wit
elliottt: The two cases from wasmtime-wasi-http that would be nice to cover are 400 and 405. We don't have anything that maps nicely to 400 right now, which is what was useful about the original `HTTP-request-error`. We were raising `405` for parse errors from hyper when trying to process the `Other` case for the `Method` variant.

acfoltzer: @elliottt what errors are you wanting to map to 400?

wit/types.wit
pchickey: @acfoltzer that is too narrow a definition of user. This spec already has implementations in progress that will be deployed beyond vendor-managed FaaS. For example the wasmtime-wasi-http implementation is intended to be run via `runwasi` in kubernetes environments, the jco implementation that could run in multiple different versions of node and node-like engines (deno, bun, etc), and there is interest in implementations that run on top of ServiceWorker, an implementation for nginx unit, and so on. In all of those cases, the user may also be managing major details of the embedding themselves, and the embeddings may encounter errors in ways that are difficult for us to anticipate here.

wit/types.wit
elliottt: When we receive an error from hyper that returns true only for `is_parse`, it's not entirely clear what we should map that to without having `HTTP-request-error` available. `400` seemed like a good use for that, but without `HTTP-request-error` I've switched back to `HTTP-protocol-error`.

acfoltzer: I think that's the right choice. The `is_parse()` predicate is pretty broad within `hyper`, but in general indicates that a successful message was not received. But `HTTP-request-error(400)` would imply that the next hop successfully returned a 400, at least going by its definition in RFC 9209

elliottt: This is the specific case I'm thinking of, where the outgoing-handler needs to reject a `Scheme::Other` value during request validation. The request hasn't been sent upstream yet, but we still need to indicate that it was malformed. After the change to remove `HTTP-request-error` there's no longer a good place for that:

https://github.com/bytecodealliance/wasmtime/pull/7434/commits/7af92d06e541f5aab7d46e9b4acb7f7df9c63be2#diff-5dc7fbb5c0f1d7e9bc2d07ee4d374589e46e4c2f054fd6d28b239da1aff426bdL55-L65

acfoltzer: Gotcha, I don't think `HTTP-request-error` would be appropriate for that case regardless, I think we need a `HTTP-request-URI-invalid` instead

pchickey:
Thanks everyone for all of the discussion here - I've mentioned this in some other contexts that we are scheduled to cut a release candidate for WASI Preview 2 at the end of this week. I would like to get this PR resolved and merged by the end of Friday Nov 10, even if it means cutting off some discussion and deferring it to a later date. @elliottt and I are going to pair his afternoon to try to resolve these outstanding threads of discussion and get this PR ready to land, and bring the wasmtime implementation up-to-date with it. cc @lukewagner @acfoltzer @PiotrSikora

wit/types.wit
acfoltzer: We're designating `internal-error` as the catch-all error case

wit/types.wit
PiotrSikora: > My rationale [laid out here](https://github.com/WebAssembly/wasi-http/pull/52#discussion_r1379277447) is that the remaining 4xx space would best be covered by returning an `ok` result with the next hop's 4xx response directly conveyed

But returning `ok` and 4xx status code would imply that the error came from the peer and not the host environment, wouldn't it?

acfoltzer: Right, the remaining 4xx codes are those that really can't be answered by the environment. If you notice a case we've missed, let's add it as new variants in a follow-up PR

PR #55, opened 2023-09-22
fix: ensure comments set as documentation

eduardomourar:
This will ensure comments are set as documentation comments. That way the description will be part of output in the different language bindgen.


pchickey:
Is it ok if we hold off on this? I have a big set of changes pending, including changing the comments to documentation, and this will just be a merge conflict.

eduardomourar:
Definitely. I just created the PR so I would not forget it.

pchickey:
This should be obsolete by #62 now.

PR #64, opened 2023-10-24
Rename the methods to retrieve `outgoing-body` from `write` to `body`

pchickey:
Prior to this PR, the `outgoing-request` and `outgoing-response` resources have methods `write` which return `result<outgoing-body>`.

`body` is a better name than `write` - it makes it explicit this is about access to a body resource. The `outgoing-body` resource itself has a method `write` which gives the body contents as a stream. Using different names for these closely related but different methods makes their distinction more clear.

Additionally: the docs for `outgoing-request.write` were incorrect - the returned `outgoing-body` is not a child of the `outgoing-request` (or `outgoing-response`). If we were to enforce a child relationship here, outgoing body streaming (i.e. writing to the body after the request or response has been initiated) would not be possible. The word `child` had been left in the docs by accident, and this PR removes it and makes the docs for both `body` methods the same.


lukewagner (approved):
lgtm!

PR #65, opened 2023-10-24
add accessor methods to outgoing-request and outgoing-response 

pchickey:
This change has two purposes:

1. adding the accessors enables treating the resource's representation of all state as canonical, instead of requiring a component to retain a copy of all state contained in the resource.

1. Adding mutators for all state supports use cases where it's required to change a request's/response's state after writing to its body. This is in particular the case when implementing preexisting systems or specifications, such as the WHATWG fetch spec used in JavaScript environments.


wit/types.wit
```diff
@@ -176,10 +228,17 @@ interface types {
...
+    /// implementation determine how to respond with an HTTP error response.
+    ///
+    /// This method may return an error when the `outgoing-response` contains
+    /// a `status-code` or `headers` which the implementation does not
+    /// support.
     set: static func(
       param: response-outparam,
       response: result<outgoing-response, error>,
-    );
+    ) -> result<_, error>;
```
lukewagner, line 241: I think we need to discuss a bit more how and when invalid headers are detected, and I think it might not be here.  However, thinking about this method more, we definitely *do* need to fail if `set` is called more than once, so how about simply returning a `result` (no payload) for now, and we can discuss the invalid header case separately.

lukewagner (approved):
Great work.  lgtm with the one requested change below:

wit/types.wit
pchickey: Its not possible to call `set` more than once because it consumes the `response-outparam`, of which only one exists. I don't think we should make it possible to retry after a failure, we just need to return a failure somehow.

We do need to make it possible for the implementation to reject an outgoing-response somewhere, whether because the status-code is not supported or other reasons. I can get rid of the mention of headers and we can settle that elsewhere.

lukewagner: Ah, good point regarding `set`-more-than-once.  However, for some of the same reasons as headers, it seems like, if there is an invalid status code, we should reject that in `set-status-code`.

pchickey: OK, I think we need to spend a bunch more time on this discussion, but for now I'm going to merge this PR because adding these accessors is orthogonal, I think.

PR #66, opened 2023-10-31
Fields, timeout, and incoming handler signature changes

elliottt:
This PR includes a number of changes that we've discussed independently, all bundled up into one:

* The `request-options` record has been turned into a resource, which allows for two new capabilities: embedders can clearly indicate when timeouts are not supported by returning `None` from a setter; the spec can evolve without breaking older guests, as new methods can be added without breaking compatibility.
* The `response-outparam.set` method has had its result type removed. In general the intent is that any validation of the status code and headers should occur before `set` is called, and that validation was the only need for the result return type.
* The `future-trailers.get` method now returns an `option<trailers>` value on success, which allows for the presence of trailers in hte request to be clearly indicated. Previously the absence of trailers was indicated by returning an empty `fields`, and now it's indicated by returning a `none` value.
* The `fields.set` and `fields.append` methods have had their return values modified to indicate potential failure, which is allowed to happen in the cases outlined by the `header-error` variant. Additionally, as we now can raise errors for header names and values, the constructor has been modified to no longer accept a list of key/value pairs. The `fields.from-list` method was added to allow for this use-case, and returns a result that can indicate if there was a problem with one of the header name/value entries.

### DIFF
--- a/wit/types.wit
+++ b/wit/types.wit
@@ -2,6 +2,7 @@
 /// HTTP Requests and Responses, both incoming and outgoing, as well as
 /// their headers, trailers, and bodies.
 interface types {
+  use wasi:clocks/monotonic-clock.{duration};
   use wasi:io/streams.{input-stream, output-stream};
   use wasi:io/poll.{pollable};
 
@@ -36,6 +37,13 @@ interface types {
     unexpected-error(string)
   }
 
+  /// This tyep enumerates the different kinds of errors that may occur when
+  /// setting or appending to a `fields` resource.
+  variant header-error {
+    invalid-syntax,
+    forbidden,
+  }
+
   /// Field keys are always strings.
   type field-key = string;
 
@@ -49,6 +57,9 @@ interface types {
   /// Headers and Trailers.
   resource fields {
 
+    /// Construct an empty HTTP Fields.
+    constructor();
+
     /// Construct an HTTP Fields.
     ///
     /// The list represents each key-value pair in the Fields. Keys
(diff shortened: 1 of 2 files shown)


elliottt:
I'm not sure why the `monotonic-clock.{duration}` use is failing, is it pulling it in from an upstream repo that hasn't been updated yet?

pchickey:
Yes, duration has not landed in wasi-clocks yet, that needs to merge and then we need to run wit-deps update.

wit/types.wit
```diff
@@ -59,22 +70,33 @@ interface types {
...
 
     /// Get all of the values corresponding to a key.
     get: func(name: field-key) -> list<field-value>;
 
     /// Set all of the values for a key. Clears any existing values for that
     /// key, if they have been set.
-    set: func(name: field-key, value: list<field-value>);
+    ///
+    /// The operation can fail if the name or value arguments are invalid, or if
+    /// the name is forbidden.
```
lukewagner, line 87: Maybe "factor out" this text describing the failure modes into the docs for the cases of `header-error` instead of repeating it here and the other 2 uses of `header-error`?

lukewagner (approved):
Looks good to me (assuming build errors re: `duration` are resolved)

wit/types.wit
elliottt: That's a good call, thanks!

pchickey:
I just merged the latest wit deps to main, so if you merge (and then regenerate your markdown) you'll have duration available.

PR #69, opened 2023-11-07
Fallible setters, and fields.delete method

elliottt:
As the setters for `outgoing-request` and `outgoing-response` take argument values that are potentially malformed, allow all setters to reject invalid inputs. This necessitated removing some constructor arguments, as wit currently does not allow for constructors to fail, replacing them instead with documented sensible defaults.

Additionally, as the `fields.delete` method is mutating the fields, change its return type to also permit a `forbidden` error.


lukewagner (approved):
lgtm

PR #70, opened 2023-11-10
all child fields are immutable, add header-error.immutable error

pchickey:
And update the doc strings throughout to specify which give a mutable fields and which give an immutable fields. also, add missed text that a future-trailers gives a child resource fields.

One of the major motivators of this change is eliminating the possibility of modifying a `Content-Length` header after an outgoing body has been created. We expect it is impossible to satisfy changes to content-length once the body is being sent, so rather than specify what the behavior of that particular problem, lets just make the fields immutable so that it is always rejected at the call to `set`, `append`, or `delete`.


lukewagner (approved):
Looks great, thanks.

PR #71, opened 2023-11-10
make outgoing-body.finish fallible in case of a Content-Length mismatch

pchickey:
`outgoing-body.finish` now fails if the body's `outgoing-request` or `outgoing-response` was constructed with a Content-Length header, and the contents written to the body (via `write`) does not match the value given in the Content-Length.


lukewagner (approved):
Makes sense

PR #72, opened 2023-11-13
set wasi:http package version to 0.2.0-rc-2023-11-10

pchickey:
and dependencies for wasi:io, clocks, random, and cli are also set to 0.2.0-rc-2023-11-10.


elliottt (approved):
:tada:

PR #73, opened 2023-11-13
WIT typo

guybedford:
I believe this is a small typo.


pchickey:
I hit merge too fast, before CI got to run! I'll follow up to fix the markdown

PR #85, opened 2023-11-27
Drop `-ms` suffix for request options

### LINKED ISSUE #78: Are timeouts on request options specified in milliseconds or nanoseconds?
[Currently](https://github.com/WebAssembly/wasi-http/blob/2befff608aa362aef20d79888b993787acd2f010/wit/types.wit#L321-L342) methods on `request-options` all end in `*-ms` which seems to imply that the argument has a unit of milliseconds. The type taken, however, is a `duration` which specifies that it's in nanoseconds. My guess is that this is an artifact of refactoring that was forgotten, but what's the intention here?

alexcrichton:
This is a leftover artifact from recent refactorings but now that it's using a `duration` type which holds its value in nanoseconds the `-ms` suffix is no longer required.

Closes #78

### DIFF
--- a/proxy.md
+++ b/proxy.md
@@ -1079,73 +1079,73 @@ another component by e.g. <code>outgoing-handler.handle</code>.</p>
 <ul>
 <li><a name="constructor_request_options.0"></a> own&lt;<a href="#request_options"><a href="#request_options"><code>request-options</code></a></a>&gt;</li>
 </ul>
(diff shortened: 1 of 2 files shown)


ricochet (approved):

elliottt:
Could you also remove the references to `milliseconds` in the `request-options` resource docs?

PR #90, opened 2023-12-04
feat: introduce `outgoing-request.into-incoming-request`

### LINKED ISSUE #89: Allow guests to construct and asynchronously write to `incoming-request`
There are use cases where invoking `incoming-handler.handle` may be desirable from within the guest, for example a Wasm HTTP server application consisting of 2 components:

1. Component A - listening and accepting connections via `wasi:sockets`, parsing the HTTP header (and optionally handling decryption if TLS is used) and passing the plaintext HTTP further down to the next component for processing
2. Component B - processing plaintext HTTP (`wasi:http/incoming-handler` implementation)

Component A would be able to correctly handle the connection, but would not be able to invoke `wasi:http/incoming-handler.handle` on Component B, because it would have no means to construct `incoming-request` required to be passed to the implementation. Moreover, even if the host would provide custom functionality to construct `incoming-request`, the guest (unlike in `wasi:http/outgoing-handler.handle` case) would not be able to asynchronously write the HTTP body bytes (and future trailers), since `wasi:http/incoming-handler.handle` signature is synchronous and calling it would block the guest

So the suggestion here is two-fold:
1. Define a constructor for `incoming-request`, or perhaps a better option would be introducing a unidirectional morphism mapping `outgoing-request` to `incoming-request`, e.g. something like `into-incoming-request: static func(this: outgoing-request) -> incoming-request` method defined on `outgoing-request` resource 
2. ~Refactor `wasi:http/incoming-handler.handle` to return an asynchronous value, just like `wasi:http/outgoing-handler.handle`, which the caller can poll for.~ After taking a better look at the existing interface, this probably would not work or would not be desired. I suppose the "correct" solution here would be to define yet another "adapter" component, which would export `wasi:http/outgoing-handler.handle` and import `wasi:http/incoming-handler.handle` and would "convert" the outgoing request into an incoming one without modifying the body, such a solution would still depend on (1.), a solution to allow body streaming downstream is TBD for this case

Do note, that a simple solution to this problem would be simply not composing such an application from two components, but rather just have one component using `wasi:sockets` and handling HTTP directly without invoking `wasi:http/incoming-handler` - while true, this diminishes the value of this package, since the "component A" with the proposed structure could be a generic, reusable HTTP server component

rvolosatovs:
Closes #89 

I wish there was a nicer way to handle this, but I am not sure what it could look like at the moment.

This does not support body streaming in a single-threaded context, but at least it allows for invoking `incoming-handler.handle`

### DIFF
--- a/proxy.md
+++ b/proxy.md
@@ -1032,6 +1032,18 @@ another component by e.g. <code>outgoing-handler.handle</code>.</p>
 <ul>
 <li><a name="method_outgoing_request.headers.0"></a> own&lt;<a href="#headers"><a href="#headers"><code>headers</code></a></a>&gt;</li>
 </ul>
+<h4><a name="static_outgoing_request.into_incoming_request"><code>[static]outgoing-request.into-incoming-request: func</code></a></h4>
+<p>Takes ownership of <a href="#outgoing_request"><code>outgoing-request</code></a>, and returns an <a href="#incoming_request"><code>incoming-request</code></a>.
+This allows guests to invoke <code>incoming-handler.handle</code> on other components.
+This function will trap if the <a href="#headers"><code>headers</code></a> child is still alive.</p>
+<h5>Params</h5>
+<ul>
+<li><a name="static_outgoing_request.into_incoming_request.this"><code>this</code></a>: own&lt;<a href="#outgoing_request"><a href="#outgoing_request"><code>outgoing-request</code></a></a>&gt;</li>
+</ul>
+<h5>Return values</h5>
+<ul>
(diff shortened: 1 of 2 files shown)


pchickey:
This issue is complicated and I want to spend time figuring out what we can do to solve it, but I don't believe that we should make this addition to the spec before releasing Preview 2 (0.2.0). Since this is additive, it is potentially a candidate for including in 0.2.1, but right now I want to hold off - sound ok?

rvolosatovs:
we won't need this in 0.3, so closing for now

PR #91, opened 2023-12-04
Add `fields.has` 

### LINKED ISSUE #82: fields.get returning an option for not found
Currently `fields.get` returns an empty list for a header that has no entry.

It might be useful to explicitly make this an option?

Alternatively, a `fields.has` might be useful for these kinds of checks as well.

A length check on a list for existence just feels a bit unintuitive even though it's semantically correct.

elliottt:
It's not currently possible to test whether a `field-key` is present in a `fields`, as our return value of an empty list could mean either that the header was missing, or had no value. Add the `fields.has` method to give an unambiguous way to test for `field-key` presence in a `fields`.

Fixes #82

### DIFF
--- a/wit/types.wit
+++ b/wit/types.wit
@@ -169,10 +169,12 @@ interface types {
       entries: list<tuple<field-key,field-value>>
     ) -> result<fields, header-error>;
 
-    /// Get all of the values corresponding to a key. Returns a `none` value
-    /// when the key isn't present, to distinguish from the case where it's
-    /// present but empty.
-    get: func(name: field-key) -> option<list<field-value>>;
+    /// Get all of the values corresponding to a key.
+    get: func(name: field-key) -> list<field-value>;
+
+    /// Returns `true` when the key is present in this `fields`. If the key is
+    /// syntactically invalid, `false` is returned.
+    has: func(name: field-key) -> bool;
 
     /// Set all of the values for a key. Clears any existing values for that
     /// key, if they have been set.
--- a/wit/types.wit
+++ b/wit/types.wit
@@ -169,7 +169,10 @@ interface types {
       entries: list<tuple<field-key,field-value>>
     ) -> result<fields, header-error>;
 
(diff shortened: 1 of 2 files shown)


lukewagner (approved):

elliottt:
After some discussion, I think the right change here is instead to add the `fields.has` function that @guybedford mentioned in #82. This addresses the immediate concern of being able to test if a header is present in the `fields`, and requires no other spec changes.

guybedford:
Could we implement both changes?

pchickey (approved):
+1 to adding `fields.has`. I appreciate that we are able to avoid changing the types on any other fields methods by using `""` to represent the empty value.

elliottt:
> Could we implement both changes?

I think that since we're using `""` to mean present-but-empty, this would mean that a return value of `[]` indicates that the key is missing. I've updated the docs on `fields.get` to indicate that `[""]` is present-but-empty, and `[]` is missing.

wit/types.wit
```diff
@@ -169,9 +169,16 @@ interface types {
       entries: list<tuple<field-key,field-value>>
     ) -> result<fields, header-error>;
 
-    /// Get all of the values corresponding to a key.
+    /// Get all of the values corresponding to a key. If the key is not present
+    /// in this `fields`, an empty list is returned. However, if the key is
+    /// present but empty, this is represented by a list with one more more
```
pchickey, line 174:
```suggestion
    /// present but empty, this is represented by a list with one or more
```

pchickey (approved):

guybedford:
This sounds reasonable in equating the empty string with an empty header value. This then makes it easy to set an empty header and compose it the other functions as @pchickey mentions in https://github.com/WebAssembly/wasi-http/pull/91#issuecomment-1839454029. In the WhatWG implementation `var h = new Headers({ a: '' }); h.append('A', ''); h.get('a');` returns `" , "`, which seems highly unnecessary so I do prefer this approach for consistency.

PR #99, opened 2024-01-31
Tweak a few `fields` docs

acfoltzer:
- This patch clarifies the behavior of the various `fields` methods when syntactically invalid inputs are given. Some methods already noted this, but `get` in particular was missing a clarification that an invalid key would result in an empty list result

- Removed some text in the doc for `from-list`. While non-UTF-8 values are not preferred for new applications, they are syntactically valid due to the [`obs-text` production](https://www.rfc-editor.org/rfc/rfc9110#section-5.5-2) in the RFC grammar.

### DIFF
--- a/wit/types.wit
+++ b/wit/types.wit
@@ -159,20 +159,18 @@ interface types {
     /// list with the same key.
     ///
     /// The tuple is a pair of the field key, represented as a string, and
-    /// Value, represented as a list of bytes. In a valid Fields, all keys
-    /// and values are valid UTF-8 strings. However, values are not always
-    /// well-formed, so they are represented as a raw list of bytes.
+    /// Value, represented as a list of bytes.
     ///
-    /// An error result will be returned if any header or value was
-    /// syntactically invalid, or if a header was forbidden.
+    /// An error result will be returned if any `field-key` or `field-value` is
+    /// syntactically invalid, or if a field is forbidden.
     from-list: static func(
       entries: list<tuple<field-key,field-value>>
     ) -> result<fields, header-error>;
 
     /// Get all of the values corresponding to a key. If the key is not present
-    /// in this `fields`, an empty list is returned. However, if the key is
-    /// present but empty, this is represented by a list with one or more
-    /// empty field-values present.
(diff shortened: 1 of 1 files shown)


wit/types.wit
```diff
@@ -159,20 +159,18 @@ interface types {
...
       entries: list<tuple<field-key,field-value>>
     ) -> result<fields, header-error>;
 
     /// Get all of the values corresponding to a key. If the key is not present
-    /// in this `fields`, an empty list is returned. However, if the key is
-    /// present but empty, this is represented by a list with one or more
-    /// empty field-values present.
+    /// in this `fields` or is syntactically invalid, an empty list is returned.
+    /// However, if the key is present but empty, this is represented by a list
+    /// with one or more empty field-values present.
```
lukewagner, line 173: This seems like the right choice given that 0.2.0 is frozen and we can't change the types.  But the next time we can make a breaking change (e.g., 0.3.0), I wonder if a syntactically-invalid key parameter should produce an error and thus the return value should be wrapped in a `result`?  The answer doesn't affect this PR, but separately we could start a list of "pending tweaks for 0.3.0" if we agree.

lukewagner (approved):
This looks right to me, nice spots!

Separately, and of interest to the WASI Subgroup process for post-0.2.0 evolution, we should discuss how these changes get published.  My inclination is that, since the changes are essentially editorial, this is still part of 0.2.0 so maybe we just update [the WASI Preview 2 snapshot](https://github.com/WebAssembly/WASI/tree/main/preview2) in-place?  (cc @pchickey @sunfishcode @ricochet )

pchickey:
With regard to pushing docs improvements up to the snapshot in the root WebAssembly/WASI repo - lets make sure that whatever editorial changes are discussed in the proposals themselves, and then I can mechanically update the version we "publish" there as needed.

wit/types.wit
acfoltzer: I think that would be more consistent with the other methods, and would support carrying it forward as a proposed change for 0.3.0.

PR #100, opened 2024-02-08
refactor the imports to its own world

Mossaka:
This PR factors out the imported APIs of http proxy to its own world, following the convention from other wasi proposals (e.g. https://github.com/WebAssembly/wasi-cli/blob/main/wit/imports.wit). 

The intention of the `wasi:http/imports` world is to be `include`d into other worlds such as `wasi:http/proxy` and `wasi:cloud-core/imports`.

Suprisingly, I wasn't expecting there to be such a huge change to the markdown files. It seems like part of the reason is that the imported interfaces are unordered?

### DIFF
--- a/wit/proxy.wit
+++ b/wit/proxy.wit
@@ -1,10 +1,8 @@
 package wasi:http@0.2.0;
 
-/// The `wasi:http/proxy` world captures a widely-implementable intersection of
-/// hosts that includes HTTP forward and reverse proxies. Components targeting
-/// this world may concurrently stream in and out any number of incoming and
-/// outgoing HTTP requests.
-world proxy {
+/// The `wasi:http/imports` world imports all the APIs for HTTP proxies.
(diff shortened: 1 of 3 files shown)


wit/proxy.wit
```diff
@@ -1,10 +1,8 @@
 package wasi:http@0.2.0;
 
-/// The `wasi:http/proxy` world captures a widely-implementable intersection of
-/// hosts that includes HTTP forward and reverse proxies. Components targeting
-/// this world may concurrently stream in and out any number of incoming and
-/// outgoing HTTP requests.
-world proxy {
+/// The `wasi:http/imports` world imports all the APIs for HTTP proxies.
+/// Is is intended to be `include`d in other worlds.
```
lukewagner, line 4:
```suggestion
/// It is intended to be `include`d in other worlds.
```

lukewagner (approved):
lgtm!

PR #101, opened 2024-02-14
[DO NOT MERGE] first draft of wasi:http@0.3.0

dicej:
This is intended to start a conversation about what we expect wasi:http@0.3.0 to look like once we have support for `stream`s, `future`s, stream `error`s, and asynchronous lifts and lowers in the component model.  In addition, I'm planning to implement a prototype of this interface using [isyswasfa](https://github.com/dicej/isyswasfa) in the near future, so I want to make sure the design is at least plausible.

High level description:

- The `incoming-handler` and `outgoing-handler` interfaces have been combined into a single `handler` interface.
- The `incoming-` and `outgoing-` variations of `request` and `response` have been combined.
- I've added a `option<request-options>` field to `request` since it would be awkward to leave it as a parameter of `handler.handle` (e.g. what would it mean to receive such a parameter for an incoming request or for a request passed from one component to the other without any use of the network?).
- I've added a `request-options-error` (analogous to `header-error`) to distinguish between unsupported fields and immutable handles.

I do not intend for this to be merged into the `main` branch of the `wasi-http` repo (since we'll presumably be doing 0.2.x work there), but I _would_ like for it to live somewhere we can iterate on it.  A few options come to mind:

- Keep it in my fork of the `wasi-http` repo for now
- Make a long-lived branch in the official repo for it
- Put it in another repo under the `WebAssembly` org (e.g. `wasi-http-0.3.0`)
- Other ideas?

### DIFF
--- a/wit/types.wit
+++ b/wit/types.wit
@@ -229,8 +229,8 @@ interface types {
     /// Construct a new `request` with a default `method` of `GET`, and
     /// `none` values for `path-with-query`, `scheme`, and `authority`.
     ///
-    /// * `body` is the contents of the body, as a stream of bytes.
     /// * `headers` is the HTTP Headers for the Response.
+    /// * `body` is the contents of the body, as a stream of bytes.
     /// * `trailers` is a `future` which resolves to the HTTP Trailers for the
     ///   Response.
     /// * `options` is optional `request-options` to be used if the request is
@@ -242,8 +242,8 @@ interface types {
     /// It is the obligation of the `handler.handle` implementation
     /// to reject invalid constructions of `request`.
     constructor(
-      body: stream<u8>,
       headers: headers,
+      body: stream<u8>,
       trailers: future<trailers>,
       options: option<request-options>
     );
@@ -367,13 +367,13 @@ interface types {
     /// different `status-code` is needed, it must be set via the
     /// `set-status-code` method.
     ///
-    /// * `body` is the contents of the body, as a stream of bytes.
     /// * `headers` is the HTTP Headers for the Response.
+    /// * `body` is the contents of the body, as a stream of bytes.
     /// * `trailers` is a `future` which resolves to the HTTP Trailers for the
     ///   Response.
     constructor(
-      body: stream<u8>,
       headers: headers,
+      body: stream<u8>,
       trailers: future<trailers>
     );
 
--- a/wit/types.wit
+++ b/wit/types.wit
@@ -2,6 +2,7 @@
 /// Requests and Responses, as well as their headers, trailers, and bodies.
 interface types {
   use wasi:clocks/monotonic-clock@0.2.0.{duration};
+  use wasi:io/error@0.2.0.{error};
 
   /// This type corresponds to HTTP standard Methods.
   variant method {
@@ -119,7 +120,7 @@ interface types {
     /// permitted because the fields are immutable.
     immutable,
   }
-  
+
   /// This type enumerates the different kinds of errors that may occur when
   /// setting fields of a `request-options` resource.
   variant request-options-error {
@@ -280,7 +281,7 @@ interface types {
     /// not a syntactically valid uri authority.
     set-authority: func(authority: option<string>) -> result;
 
-    /// Set the `request-options` to be associated with this request
+    /// Get the `request-options` to be associated with this request
     ///
     /// The returned `request-options` resource is immutable: `set-*` operations
     /// will fail if invoked.
@@ -288,7 +289,7 @@ interface types {
     /// This `request-options` resource is a child: it must be dropped before
     /// the parent `request` is dropped, or its ownership is transfered to
     /// another component by e.g. `handler.handle`.
-    set-options: func(options: option<request-options>);
+    options: func() -> option<request-options>;
 
     /// Get the headers associated with the Request.
     ///
--- a/wit/types.wit
+++ b/wit/types.wit
@@ -231,8 +231,8 @@ interface types {
     ///
     /// * `headers` is the HTTP Headers for the Response.
     /// * `body` is the contents of the body, as a stream of bytes.
-    /// * `trailers` is a `future` which resolves to the HTTP Trailers for the
-    ///   Response.
+    /// * `trailers` is an optional `future` which resolves to the HTTP Trailers
+    ///   for the Response.
     /// * `options` is optional `request-options` to be used if the request is
     ///   sent over a network connection.
     ///
@@ -244,7 +244,7 @@ interface types {
     constructor(
       headers: headers,
       body: stream<u8>,
-      trailers: future<trailers>,
+      trailers: option<future<trailers>>,
       options: option<request-options>
     );
 
@@ -318,14 +318,14 @@ interface types {
 
     /// Takes ownership of `request`, and returns a `trailers`.  This function will
     /// trap if a `stream` child is still alive.
-    finish: static func(this: request) -> trailers;
+    finish: static func(this: request) -> result<option<trailers>, error-code>;
   }
 
   /// Parameters for making an HTTP Request. Each of these parameters is
   /// currently an optional timeout applicable to the transport layer of the
   /// HTTP protocol.
   ///
-  /// These timeouts are separate from any the user may use to bound a an
+  /// These timeouts are separate from any the user may use to bound an
   /// asynchronous call.
   resource request-options {
     /// Construct a default `request-options` value.
@@ -369,12 +369,12 @@ interface types {
     ///
     /// * `headers` is the HTTP Headers for the Response.
     /// * `body` is the contents of the body, as a stream of bytes.
-    /// * `trailers` is a `future` which resolves to the HTTP Trailers for the
-    ///   Response.
+    /// * `trailers` is an optional `future` which resolves to the HTTP Trailers
+    ///   for the Response.
     constructor(
       headers: headers,
       body: stream<u8>,
-      trailers: future<trailers>
+      trailers: option<future<trailers>>
     );
 
     /// Get the HTTP Status Code for the Response.
@@ -412,6 +412,6 @@ interface types {
 
     /// Takes ownership of `response`, and returns a `trailers`.  This function will
     /// trap if a `stream` child is still alive.
-    finish: static func(this: response) -> trailers;
+    finish: static func(this: response) -> result<option<trailers>, error-code>;
   }
 }
--- a/wit/types.wit
+++ b/wit/types.wit
@@ -1,11 +1,7 @@
-/// This interface defines all of the types and methods for implementing
-/// HTTP Requests and Responses, both incoming and outgoing, as well as
-/// their headers, trailers, and bodies.
+/// This interface defines all of the types and methods for implementing HTTP
+/// Requests and Responses, as well as their headers, trailers, and bodies.
 interface types {
   use wasi:clocks/monotonic-clock@0.2.0.{duration};
-  use wasi:io/streams@0.2.0.{input-stream, output-stream};
-  use wasi:io/error@0.2.0.{error as io-error};
-  use wasi:io/poll@0.2.0.{pollable};
 
   /// This type corresponds to HTTP standard Methods.
   variant method {
@@ -95,18 +91,17 @@ interface types {
     field-size: option<u32>
   }
 
-  /// Attempts to extract a http-related `error` from the wasi:io `error`
+  /// Attempts to extract a http-related `error-code` from the stream `error`
   /// provided.
   ///
-  /// Stream operations which return
-  /// `wasi:io/stream/stream-error::last-operation-failed` have a payload of
-  /// type `wasi:io/error/error` with more information about the operation
-  /// that failed. This payload can be passed through to this function to see
-  /// if there's http-related information about the error to return.
+  /// Stream operations may fail with a stream `error` with more information
+  /// about the operation that failed. This `error` can be passed to this
+  /// function to see if there's http-related information about the error to
+  /// return.
   ///
-  /// Note that this function is fallible because not all io-errors are
+  /// Note that this function is fallible because not all stream errors are
   /// http-related errors.
-  http-error-code: func(err: borrow<io-error>) -> option<error-code>;
+  http-error-code: func(err: borrow<error>) -> option<error-code>;
 
   /// This type enumerates the different kinds of errors that may occur when
   /// setting or appending to a `fields` resource.
@@ -122,6 +117,17 @@ interface types {
 
     /// This error indicates that the operation on the `fields` was not
     /// permitted because the fields are immutable.
+    immutable,
+  }
+  
+  /// This type enumerates the different kinds of errors that may occur when
+  /// setting fields of a `request-options` resource.
+  variant request-options-error {
+    /// Indicates the specified field is not supported by this implementation.
+    not-supported,
+
+    /// Indicates that the operation on the `request-options` was not permitted
+    /// because it is immutable.
     immutable,
   }
 
@@ -140,9 +146,9 @@ interface types {
   /// A `fields` may be mutable or immutable. A `fields` created using the
   /// constructor, `from-list`, or `clone` will be mutable, but a `fields`
   /// resource given by other means (including, but not limited to,
-  /// `incoming-request.headers`, `outgoing-request.headers`) might be be
-  /// immutable. In an immutable fields, the `set`, `append`, and `delete`
-  /// operations will fail with `header-error.immutable`.
+  /// `request.headers`) might be be immutable. In an immutable fields, the
+  /// `set`, `append`, and `delete` operations will fail with
+  /// `header-error.immutable`.
   resource fields {
 
     /// Construct an empty HTTP Fields.
@@ -217,60 +223,32 @@ interface types {
   /// Trailers is an alias for Fields.
   type trailers = fields;
 
-  /// Represents an incoming HTTP Request.
-  resource incoming-request {
+  /// Represents an HTTP Request.
+  resource request {
 
-    /// Returns the method of the incoming request.
-    method: func() -> method;
-
-    /// Returns the path with query parameters from the request, as a string.
-    path-with-query: func() -> option<string>;
-
-    /// Returns the protocol scheme from the request.
-    scheme: func() -> option<scheme>;
-
-    /// Returns the authority from the request, if it was present.
-    authority: func() -> option<string>;
-
-    /// Get the `headers` associated with the request.
-    ///
-    /// The returned `headers` resource is immutable: `set`, `append`, and
-    /// `delete` operations will fail with `header-error.immutable`.
-    ///
-    /// The `headers` returned are a child resource: it must be dropped before
-    /// the parent `incoming-request` is dropped. Dropping this
-    /// `incoming-request` before all children are dropped will trap.
-    headers: func() -> headers;
-
-    /// Gives the `incoming-body` associated with this request. Will only
-    /// return success at most once, and subsequent calls will return error.
-    consume: func() -> result<incoming-body>;
-  }
-
-  /// Represents an outgoing HTTP Request.
-  resource outgoing-request {
-
-    /// Construct a new `outgoing-request` with a default `method` of `GET`, and
+    /// Construct a new `request` with a default `method` of `GET`, and
     /// `none` values for `path-with-query`, `scheme`, and `authority`.
     ///
     /// * `headers` is the HTTP Headers for the Request.
     ///
     /// It is possible to construct, or manipulate with the accessor functions
-    /// below, an `outgoing-request` with an invalid combination of `scheme`
+    /// below, an `request` with an invalid combination of `scheme`
     /// and `authority`, or `headers` which are not permitted to be sent.
-    /// It is the obligation of the `outgoing-handler.handle` implementation
-    /// to reject invalid constructions of `outgoing-request`.
+    /// It is the obligation of the `handler.handle` implementation
+    /// to reject invalid constructions of `request`.
     constructor(
-      headers: headers
+      body: body,
+      headers: headers,
+      trailers: option<future<trailers>>,
+      options: option<request-options>
     );
 
-    /// Returns the resource corresponding to the outgoing Body for this
-    /// Request.
+    /// Returns the resource corresponding to the Body for this Request.
     ///
-    /// Returns success on the first call: the `outgoing-body` resource for
-    /// this `outgoing-request` can be retrieved at most once. Subsequent
-    /// calls will return error.
-    body: func() -> result<outgoing-body>;
+    /// Returns success on the first call: the `body` resource for this
+    /// `request` can be retrieved at most once. Subsequent calls will return
+    /// error.
+    body: func() -> result<body>;
 
     /// Get the Method for the Request.
     method: func() -> method;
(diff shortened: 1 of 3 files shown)


wit/types.wit
```diff
@@ -483,88 +400,15 @@ interface types {
...
-    /// status and headers have recieved successfully, or that an error
-    /// occured. Errors may also occur while consuming the response body,
-    /// but those will be reported by the `incoming-body` and its
-    /// `output-stream` child.
-    get: func() -> option<result<result<incoming-response, error-code>>>;
-
+    /// Returns success on the first call: the `body` resource for this
+    /// `response` can be retrieved at most once. Subsequent calls will return
+    /// error.
+    body: func() -> result<body>;
```
lukewagner, line 412: I wonder if, for the same reason that we have a `static finish` that consumes the body to produce the trailers (thereby avoiding leaving `body` resources in an indeterminate state), if we should have a `static consume` on `request`/`response` that consumes the `request`/`response` (thereby avoiding the "exists, but doesn't have a valid body, so probably don't hand off this request/response to anyone" state).  This would be complemented by a `take-headers` method that lets you hold onto the headers even after consuming the request/response.

Once you make that change in your head, I think we can simplify one step further and have, directly on both `request` and `response`:
* `consume: func() -> result<stream<u8>>`
* `static finish: func(request|response) -> future<trailers>`

Lastly, because with streams it is clear exactly how many elements are successfully passed from producer to consumer, we could even remove the "can only call `consume` once rule" and say that if you call `consume` and drop the stream, you can call `consume` again and pick up where you left off (repeatedly up until you call `finish`).  There would still need to be a `result` to fail if you call `consume` concurrent with a previous call to `consume`.  I think this might even match the semantics of Web Streams, which have a [locked](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream/locked) state that is set when you create a stream "reader" and can be unlocked by dropping that reader (and thus our `stream` handle that can be dropped by the caller is rather like the Web Stream reader).  Overall, what this would allow is for component C1 to be given a request, to consume some bytes, and then pass on the request to C2, which consumes some more.

WDYT?

lukewagner:
Very exciting to see these simplifications, even if only hypothetical atm.

wit/types.wit
dicej: SGTM!

wit/types.wit
```diff
@@ -217,72 +223,44 @@ interface types {
...
+    /// below, an `request` with an invalid combination of `scheme`
     /// and `authority`, or `headers` which are not permitted to be sent.
-    /// It is the obligation of the `outgoing-handler.handle` implementation
-    /// to reject invalid constructions of `outgoing-request`.
+    /// It is the obligation of the `handler.handle` implementation
+    /// to reject invalid constructions of `request`.
     constructor(
-      headers: headers
+      body: body,
+      headers: headers,
```
rylev: nit: is there a reason for this particular order of params? It's not a big deal, but it would seem more consistent to take `headers` first since that's how they're laid out on the wire.

wit/types.wit
```diff
@@ -329,146 +317,75 @@ interface types {
...
-    /// If a different `status-code` is needed, it must be set via the
+    /// Construct an `response`, with a default `status-code` of `200`.  If a
+    /// different `status-code` is needed, it must be set via the
     /// `set-status-code` method.
     ///
     /// * `headers` is the HTTP Headers for the Response.
-    constructor(headers: headers);
+    constructor(
+      body: body,
+      headers: headers,
```
rylev, line 376: same nit as above.

wit/types.wit
dicej: I just pushed an update with those changes.

wit/types.wit
```diff
@@ -217,72 +223,42 @@ interface types {
...
     /// and `authority`, or `headers` which are not permitted to be sent.
-    /// It is the obligation of the `outgoing-handler.handle` implementation
-    /// to reject invalid constructions of `outgoing-request`.
+    /// It is the obligation of the `handler.handle` implementation
+    /// to reject invalid constructions of `request`.
     constructor(
-      headers: headers
+      body: stream<u8>,
+      headers: headers,
+      trailers: future<trailers>,
```
lukewagner, line 247:
```suggestion
      trailers: option<future<trailers>>,
```

wit/types.wit
```diff
@@ -304,23 +280,53 @@ interface types {
...
+    /// to not inhibit delivery of the trailers if the user does not read the
+    /// entire body.
+    ///
+    /// This function may be called multiple times as long as any `stream`s
+    /// returned by previous calls have been dropped first.
+    body: func() -> result<stream<u8>>;
+
+    /// Takes ownership of `request`, and returns a `trailers`.  This function will
+    /// trap if a `stream` child is still alive.
+    finish: static func(this: request) -> trailers;
```
lukewagner, line 321:
```suggestion
    finish: static func(this: request) -> result<option<trailers>, error-code>;
```

wit/types.wit
```diff
@@ -304,23 +280,53 @@ interface types {
...
   }
 
   /// Parameters for making an HTTP Request. Each of these parameters is
   /// currently an optional timeout applicable to the transport layer of the
   /// HTTP protocol.
   ///
-  /// These timeouts are separate from any the user may use to bound a
-  /// blocking call to `wasi:io/poll.poll`.
+  /// These timeouts are separate from any the user may use to bound a an
```
lukewagner, line 328:
```suggestion
  /// These timeouts are separate from any the user may use to bound an
```

wit/types.wit
```diff
@@ -329,146 +335,47 @@ interface types {
...
     ///
+    /// * `body` is the contents of the body, as a stream of bytes.
     /// * `headers` is the HTTP Headers for the Response.
-    constructor(headers: headers);
+    /// * `trailers` is a `future` which resolves to the HTTP Trailers for the
+    ///   Response.
+    constructor(
+      body: stream<u8>,
+      headers: headers,
+      trailers: future<trailers>
```
lukewagner, line 377:
```suggestion
      trailers: option<future<trailers>>
```

wit/types.wit
```diff
@@ -483,88 +390,28 @@ interface types {
...
-    /// but those will be reported by the `incoming-body` and its
-    /// `output-stream` child.
-    get: func() -> option<result<result<incoming-response, error-code>>>;
+    /// This function may be called multiple times as long as any `stream`s
+    /// returned by previous calls have been dropped first.
+    body: func() -> result<stream<u8>>;
 
+    /// Takes ownership of `response`, and returns a `trailers`.  This function will
+    /// trap if a `stream` child is still alive.
+    finish: static func(this: response) -> trailers;
```
lukewagner, line 415:
```suggestion
    finish: static func(this: response) -> result<option<trailers>, error-code>;
```

lukewagner (commented):
Nice, looks good so far!

wit/types.wit
dicej: Yeah, I originally wrapped it in an `option`, then removed it, thinking it should be more symmetric with the `headers` parameter; i.e. if you have no headers, just pass an empty `headers`, and if you have no trailers (or don't know yet whether you'll have any), you pass a `future` that (may) resolve to an empty `trailers`.  In other words, if we make `trailers` optional, should we also make `headers` optional?

lukewagner: Fair point and I wondered the same.  Here, I was mostly just matching 0.2 (which made trailers optional, but headers non-optional).  But perhaps you could make the argument that you almost always want headers but trailers are pretty rare, but maybe for symmetry both should be optional.

dicej: Yeah, trailers _are_ rare, and forcing everyone to create and pass a future would be annoying.  I'll go ahead and make them optional and leave the headers required for now.

dicej: We might even want to remove the `trailers` param from the constructor and add a setter instead.  How does that sound?

lukewagner: There's an interesting consequence to that: passing a `future` to a method (as opposed to ctor) holds onto the borrowed self-request handle passed to the method-call until the future is resolved which in turn prevents you from transferring ownership of the request out of the current component (which one naturally wants to do before the trailers are resolved).  The ctor doesn't have this problem b/c the ctor doesn't borrow anything.

(As a side note: in most cases, this borrow-lasts-the-whole-async-call rule makes sense (e.g., I shouldn't be transferring ownership of a file in the middle of doing a streaming `write` method call), but in this case it maybe seems restrictive.  If this is a problem, there are various ways to avoid this limitation (in the design or using an intermediate constructed resource type) that we can discuss.)

dicej: Yeah, I was wondering about the borrow.  Rust `async` functions have the same property, which you can work around by writing a non-async function that returns a `Future` with a `'static` lifetime, but that trick won't work in our "colorless" approach (which is fine).

Anyway, let's just leave it as a constructor param for now.

lukewagner: Interesting!  I suppose the "intermediate constructed resource type" workaround is somewhat analogous to the `'static` workaround.  That is, if we really didn't want a constructor parameter, we could have:
```wit
resource future-trailers {
  constructor(f: future<trailers>);
}
resource request {
  set-trailers: func(f: future-trailers);
}
```
If this became a necessary or common workaround, I wonder if we couldn't come up with some more first-class support for this technique that avoided the superfluous resource type.  But maybe the constructor-parameter approach ends up being good enough for most cases.  I guess we'll have to see what happens.

dicej: Would something like this work?
```
resource request {
  constructor(headers: headers, body: stream<u8>);
  set-trailers: static func(request: request, trailers: future<trailers>) -> request;
  ...
}
```

dicej:
I've just finished adding a "mock" implementation of this draft (i.e. one suitable for testing but which doesn't use the network) to [isyswasfa](https://github.com/dicej/isyswasfa).  The [service](https://github.com/dicej/isyswasfa/blob/main/test/service/src/lib.rs) and [middleware](https://github.com/dicej/isyswasfa/blob/main/test/middleware/src/lib.rs) examples illustrate the structure of a non-trivial application that includes body streaming and trailers.  In other words, that's roughly what I would expect a real `wasi:http@0.3.0` Rust app to look like modulo the naming of the "isyswasfa" pieces, which would instead be canon or WIT built-ins in the real version.

A few observations based on the implementation experience:

- Now that there's a single `wasi:http/handler` interface, if middleware wants to both forward requests to another component _and_ make outbound requests (or wants to be able to forward requests to multiple distinct components), we'll need some way for it to import multiple `wasi:http/hander`-like things.  This has also come up in the context of virtualization, and AFAIK all proposed solutions are currently theoretical and require e.g. WIT changes.
-  It would be nice if we had a `take-trailers: func() -> option<future<trailers>>` method on both `request` and `response`, which would allow middleware to forward trailers to a new `request/response` without having to pipe the body as well (see the middleware example above where we're forced to do this).  Either that, or allow mutating the headers of a `request` or `response` so we don't have to create a new one from scratch (which involves copying everything over, including trailers) when all we want to do is add/remove/edit headers.  Thoughts or concerns?
- In practice, it would be great to have WIT-level "hints" about whether a binding generator should generate async or sync bindings for each function.  For example, getters and setters on resources probably shouldn't be async by default, whereas e.g. `request.finish` definitely should be.  And when I say "default", I mean that the app developer can always override it and say "no, really, I want to make async calls to `request.method`" or "no really, I want to make sync calls to `request.finish`", or even "please generate both a sync and an async version for this imported function".  The hints would only affect the defaults for the bindings generator without creating any sort of "function coloring" at the component type level.
    - Currently, I'm hard-coding which `wasi:http` functions get sync bindings and which ones get async bindings, which is awkward.

lukewagner:
Great work!  That's really exciting.

> we'll need some way for it to import multiple wasi:http/hander-like things

Yes, that makes sense.  I think the WIT feature we need to add here is [CM/#287](https://github.com/WebAssembly/component-model/issues/287), so you'd be able to write:
```wit
world middleware {
  import a: wasi:http/handler;
  import b: wasi:http/handler;
  ...
}
```
which would then produce a component-type:
```wat
(component
  (import "a" (implements "wasi:http/handler") (instance ...))
  (import "b" (implements "wasi:http/handler") (instance ...))
  ...
)
```
where `(implements ...)` is neither part of the name nor the type; it's just a semantic hint (that enables WIT roundtripping but also perhaps improved DX in various composition scenarios).

> It would be nice if we had a take-trailers: func() -> option<future<trailers>> method on both request and response, which would allow middleware to forward trailers to a new request/response without having to pipe the body as well (see the middleware example above where we're forced to do this).

So just to check my understanding: is the problem that, if you immediately call `finish` on the body, it doesn't give you an `option<future<trailers>>`, instead you get an async call that you can wait on to get an `option<trailers>` (no `future`)?  I wonder if the fix here is to have some built-in way to get a `future<T*>` from any lowered async call to a function returning `T*`.  I could imagine a bunch of different ABI variations, not sure which one is best, but it seems imminently doable. 

> Either that, or allow mutating the headers of a request or response so we don't have to create a new one from scratch (which involves copying everything over, including trailers) when all we want to do is add/remove/edit headers.

#103 is proposes adding a `take-headers` for extracting headers that become mutable (once they are disconnected from the `request`/`response`).  Are you taking that addition into account and still it's insufficient?  (Alternatively, it seems like we could just make all fields other than `Content-Length` mutable; the current rules are just trying to be regular, but `Content-Length` is the only one that really wants the current rules.)

> In practice, it would be great to have WIT-level "hints" about whether a binding generator should generate async or sync bindings for each function.

Yes, agreed.  I was thinking of adding a `nonblocking` function attribute (so, e.g., you could have `foo: nonblocking func(s: string) -> string`) that would not just hint, but actually enforce the invariant that the callee did not transitively call a blocking import.  I think it's important for this not to be the default (b/c it really hinders what's possible w.r.t virtualizability) so that interface designers think about it.  But the WIT syntactic sugar for: constructors, getters and setters could set it by default (b/c for many languages, if the bindings generator has to worry about async, it won't be able to actually generate the bindings you want it to generate because they are all synchronous).

dicej:
> > we'll need some way for it to import multiple wasi:http/hander-like things
> 
> Yes, that makes sense. I think the WIT feature we need to add here is [CM/#287](https://github.com/WebAssembly/component-model/issues/287)

Yeah, that looks great.

> > It would be nice if we had a take-trailers: func() -> option<future> method on both request and response, which would allow middleware to forward trailers to a new request/response without having to pipe the body as well (see the middleware example above where we're forced to do this).
> 
> So just to check my understanding: is the problem that, if you immediately call `finish` on the body, it doesn't give you an `option<future<trailers>>`, instead you get an async call that you can wait on to get an `option<trailers>` (no `future`)? I wonder if the fix here is to have some built-in way to get a `future<T*>` from any lowered async call to a function returning `T*`. I could imagine a bunch of different ABI variations, not sure which one is best, but it seems imminently doable.

In the current draft, the `finish` method is on `request`/`response`, and it is `static` and consumes its argument, which means any child handles must have already been dropped, including the body.  So if I want to take a request and make a new request that's exactly the same as the old one (same body, trailers, etc.) but with different headers, I _must_ spawn a task that copies the body buffer-by-buffer, _then_ call `finish`, get the trailers, and send them on to the new request.  In other words, I can't just take the body from the first request and give it to the second request and then call `finish` on the first request, because that will trap.  I'm not sure how being able to get a `future<T>` from a lowered async call would help in this case -- would the returned `future<T>` somehow own the request passed to `finish` and keep it alive so that the body can stay alive, too?

> > Either that, or allow mutating the headers of a request or response so we don't have to create a new one from scratch (which involves copying everything over, including trailers) when all we want to do is add/remove/edit headers.
> 
> #103 is proposes adding a `take-headers` for extracting headers that become mutable (once they are disconnected from the `request`/`response`). Are you taking that addition into account and still it's insufficient? (Alternatively, it seems like we could just make all fields other than `Content-Length` mutable; the current rules are just trying to be regular, but `Content-Length` is the only one that really wants the current rules.)

If I want to pass the original request to the imported `wasi:http/handler` instead of creating a brand new request and passing that, the only way I can modify headers is if I get mutable access to the headers of the original request (i.e. not a disconnected version), so yes, I think we'd need the mutable-except-for-content-length approach.

> > In practice, it would be great to have WIT-level "hints" about whether a binding generator should generate async or sync bindings for each function.
> 
> Yes, agreed. I was thinking of adding a `nonblocking` function attribute (so, e.g., you could have `foo: nonblocking func(s: string) -> string`) that would not just hint, but actually enforce the invariant that the callee did not transitively call a blocking import. I think it's important for this not to be the default (b/c it really hinders what's possible w.r.t virtualizability) so that interface designers think about it. But the WIT syntactic sugar for: constructors, getters and setters could set it by default (b/c for many languages, if the bindings generator has to worry about async, it won't be able to actually generate the bindings you want it to generate because they are all synchronous).

:+1:

lukewagner:
> [...] so yes, I think we'd need the mutable-except-for-content-length approach.

Good point; I think you're right.  This isn't clear-cut in 0.2 because you end up being forced to create a new resource anyways, but in 0.3, it seems clear.

>  So if I want to take a request and make a new request that's exactly the same as the old one (same body, trailers, etc.) but with different headers

I think the mutable-except-for-content-length design fixes this exact issue, but if we tweak it slightly (maybe I want to create new headers but use the same body), it does seem like a design flaw if you have to wait to copy over the body stream just so you can forward the trailers.  I know I proposed removing the intermediate `body` resource type, but I think this is a good argument for adding it back again.  Because then you could have `consume: func() -> result<body>` and then in one simple transfer of ownership, you could pass it on to a new request or response.  Does that sound right to you?

dicej:
> I know I proposed removing the intermediate `body` resource type, but I think this is a good argument for adding it back again. Because then you could have `consume: func() -> result<body>` and then in one simple transfer of ownership, you could pass it on to a new request or response. Does that sound right to you?

Yeah, that should work.  If we do that and leave the `finish` static methods as they are, then we'd also need the "async-call-to-`future<T>`" feature so we can call `finish` without awaiting it and pass the call as a `future<trailers>` to the constructor for the new request or response.

lukewagner:
Agreed!

dicej:
Does this look plausible?

```diff
diff --git a/wit/deps/http/types.wit b/wit/deps/http/types.wit
index c859a38..9793aca 100644
--- a/wit/deps/http/types.wit
+++ b/wit/deps/http/types.wit
@@ -225,6 +225,43 @@ interface types {
   /// Trailers is an alias for Fields.
   type trailers = fields;
 
+  /// Represents an HTTP Request or Response's Body.
+  ///
+  /// A body has both its contents - a stream of bytes - and a (possibly empty)
+  /// set of trailers, indicating that the full contents of the body have been
+  /// received. This resource represents the contents as a `stream<u8>` and the
+  /// delivery of trailers as a `trailers`, and ensures that the user of this
+  /// interface may only be consuming either the body contents or waiting on
+  /// trailers at any given time.
+  resource body {
+  
+    /// Construct a new `body` with the specified content and trailers.
+    constructor(
+      body: stream<u8>,
+      trailers: option<future<trailers>>
+    );
+  
+    /// Returns the contents of the body, as a stream of bytes.
+    ///
+    /// The returned `stream` is a child: it must be dropped before the parent
+    /// `body` is dropped, or consumed by `body.finish`.
+    ///
+    /// This invariant ensures that the implementation can determine whether the
+    /// user is consuming the contents of the body, waiting on the trailers to
+    /// be ready, or neither. This allows for network backpressure is to be
+    /// applied when the user is consuming the body, and for that backpressure
+    /// to not inhibit delivery of the trailers if the user does not read the
+    /// entire body.
+    ///
+    /// This function may be called multiple times as long as any `stream`s
+    /// returned by previous calls have been dropped first.
+    body: func() -> result<stream<u8>>;
+
+    /// Takes ownership of `body`, and returns a `trailers`.  This function will
+    /// trap if a `stream` child is still alive.
+    finish: static func(this: response) -> result<option<trailers>, error-code>;
+  }
+  
   /// Represents an HTTP Request.
   resource request {
 
@@ -245,8 +282,7 @@ interface types {
     /// to reject invalid constructions of `request`.
     constructor(
       headers: headers,
-      body: stream<u8>,
-      trailers: option<future<trailers>>,
+      body: body,
       options: option<request-options>
     );
 
@@ -302,25 +338,8 @@ interface types {
     /// component by e.g. `handler.handle`.
     headers: func() -> headers;
 
-    /// Returns the contents of the body, as a stream of bytes.
-    ///
-    /// The returned `stream` is a child: it must be dropped before the parent
-    /// `request` is dropped, or consumed by `request.finish`.
-    ///
-    /// This invariant ensures that the implementation can determine whether the
-    /// user is consuming the contents of the body, waiting on the trailers to
-    /// be ready, or neither. This allows for network backpressure is to be
-    /// applied when the user is consuming the body, and for that backpressure
-    /// to not inhibit delivery of the trailers if the user does not read the
-    /// entire body.
-    ///
-    /// This function may be called multiple times as long as any `stream`s
-    /// returned by previous calls have been dropped first.
-    body: func() -> result<stream<u8>>;
-
-    /// Takes ownership of `request`, and returns a `trailers`.  This function will
-    /// trap if a `stream` child is still alive.
-    finish: static func(this: request) -> result<option<trailers>, error-code>;
+    /// Takes ownership of the `request` and returns the `body`.
+    consume: static func(this: request) -> body;
   }
 
   /// Parameters for making an HTTP Request. Each of these parameters is
@@ -375,8 +394,7 @@ interface types {
     ///   for the Response.
     constructor(
       headers: headers,
-      body: stream<u8>,
-      trailers: option<future<trailers>>
+      body: body,
     );
 
     /// Get the HTTP Status Code for the Response.
@@ -396,24 +414,7 @@ interface types {
     /// component by e.g. `handler.handle`.
     headers: func() -> headers;
 
-    /// Returns the contents of the body, as a stream of bytes.
-    ///
-    /// The returned `stream` is a child: it must be dropped before the parent
-    /// `response` is dropped, or consumed by `response.finish`.
-    ///
-    /// This invariant ensures that the implementation can determine whether the
-    /// user is consuming the contents of the body, waiting on the trailers to
-    /// be ready, or neither. This allows for network backpressure is to be
-    /// applied when the user is consuming the body, and for that backpressure
-    /// to not inhibit delivery of the trailers if the user does not read the
-    /// entire body.
-    ///
-    /// This function may be called multiple times as long as any `stream`s
-    /// returned by previous calls have been dropped first.
-    body: func() -> result<stream<u8>>;
-
-    /// Takes ownership of `response`, and returns a `trailers`.  This function will
-    /// trap if a `stream` child is still alive.
-    finish: static func(this: response) -> result<option<trailers>, error-code>;
+    /// Takes ownership of the `response` and returns the `body`.
+    consume: static func(this: request) -> body;
   }
 }
```

lukewagner:
@dicej Looks generally in the right direction!  A few thoughts:

First, the comments about the `body.stream` and `body.finish` functions are correct, but they should be automatically implied simply by the C-M rules of `stream` (and thus not something that has to be manually mentioned for each WASI interface or manually implemented (it would be enforced as part of the C-M bindings).  In particular, the call to `body.stream` will not return until the returned `stream` is dropped, and the ongoing `body.stream` call will hold onto the `borrow` of the `body` until then, and thus it won't be possible to `drop` or `finish` the `body` while there is any active child stream.  (If you want to keep the comments just so new-comers understand the rules, that's totally fine; maybe just add a note that this is an automatic consequence of the C-M.)

Next, for the case where I want to simply read the body bytes of a `request` or `response` into my component, I wonder if, instead of having `static` `consume` functions that take ownership of the parent `request`/`response`, what if we instead had `body` methods that return a child handle to the `body` resource (just like the `headers` method returns a child handle) with the same child rules (you have to drop any handle to the `body` before dropping or transferring ownership of the parent `request`/`response`).  This allows a component A to be given a `request`, read a few bytes of the body, close the stream, drop the `body` handle, then transfer ownership of the `request` to component B which can do likewise.  This scenario might arise concretely with, e.g., [request batching](https://api7.ai/blog/batch-request-processing-with-api-gateway).

For the case where I want to extract the body from one `request`/`response` to use it to zero-copy construct another `request`/`response`, I wonder if we should also consider the `headers` and the `into-parts` idea Adam initially proposed in #103 (that consumes the parent `request`/`response` and returns all its contents).  Because you can just call getters to access the value fields (like the `method`), I would think `into-parts` would only return the owned child resources (`headers` and `body`).  What's neat is that the child rules ensure that we don't have to worry about dangling child `headers`s, `body`s and `stream`s.

WDYT?

dicej:
All sounds good to me.

Now I'm thinking we should aim to get this merged into the repo so we can iterate on it properly (without interfering with the 0.2.x work).  Should I add a `wit-0.3.0` directory and put it there, perhaps?

lukewagner:
Sounds good to me; maybe `wit-0.3.0-draft`?  (Btw, I could be wrong, but I don't think we need dates in the WIT `package` versions, just `-draft` would be fine (unless you want them for your own experimental prototype purposes.)

dicej:
I've opened a new PR that puts the draft in its own directory: https://github.com/WebAssembly/wasi-http/pull/106

PR #103, opened 2024-02-17
Add static `into-parts` for incoming messages; deprecate `consume`

### LINKED ISSUE #102: Methods needed to get owned headers out of requests and responses
The main case where this is currently painful is when trying to perform modifications while forwarding along an `incoming-request` as an `outgoing-request`, or `incoming-response` as an `outgoing-response`. The desired workflow is something like:

```
headers <- incoming.headers();
headers.delete("header-I-want-to-filter");
outgoing <- outgoing-request.new(headers);
```

Currently, this fails because `headers` is a child resource of `incoming_request` where the underlying headers are immutable, so methods like `append` and `delete` are not permitted. Invoking `outgoing-request.new` with this child resource is fine, with embedders able to clone the headers under the hood. But once it's in the `outgoing` resource it's once again only accessible via an immutable child resource.

As far as I can tell, the only way to modify headers in this proxying scenario is to explicitly clone the headers:

```
headers <- incoming.headers();
new_headers <- headers.clone();
new_headers.delete("header-I-want-to-filter");
outgoing <- outgoing-request.new(new_headers);
```

This isn't the end of the world, but it would be nice to avoid this clone _and_ the implicit clone performed by constructing `outgoing-request` with a child resource by offering a means to get the headers back from these types as an owned resource. Perhaps we could, similar to Rust's `http::Request::into_parts()` method, augment the `consume` method to also return the headers (and other parts like method, authority, path, status code [for responses]) along with the body. Strawman signatures:

```
resource incoming-request {
  consume: func() -> result<(
    method,
    option<scheme>,
    option<string>, // authority
    option<string>, // path-with-query
    headers,
    incoming-body
  )>;
}

resource incoming-response {
  consume: func() -> result<(status-code, headers, incoming-body)>;
}
```

This will remain relevant even once 0.3.0 arrives with the unification of the `incoming-` and `outgoing-` types. If the headers can still only be accessed immutably from an existing value, proxy applications wishing to perform modifications will still have to get a mutable headers resource via some means, and absent other changes that'll mean an unnecessary clone.

acfoltzer:
Closes #102. See discussion in that issue for the motivation and design of this change.


wit/types.wit
```diff
@@ -244,7 +244,23 @@ interface types {
 
     /// Gives the `incoming-body` associated with this request. Will only
     /// return success at most once, and subsequent calls will return error.
+    ///
+    /// Deprecated in favor of `into-parts`.
```
lukewagner, line 248: I'm not sure we need to mark `consume` as *deprecated* in 0.2.1  -- if all you want is to do is read the body stream, it's a reasonable method to call (and perhaps a bit more efficient; not forcing you to copy strings you don't want).  Note that 0.3 is free and expected to make big breaking changes, and if you check out #101, you can see a sketch of how it might look.

wit/types.wit
```diff
@@ -244,7 +244,23 @@ interface types {
...
+    into-parts: static func(
+      response: incoming-request
+    ) -> tuple<
+      method,
+      option<string>, // path-with-query
+      option<scheme>,
+      option<string>, // authority
+      headers,
+      incoming-body
+    >;
```
lukewagner, line 263: Given that one can already copy everything except the `headers` and `incoming-body` fields, I wonder if we really need a method that copies all fields at once since it may often be the case that a number of the fields will go unused (which won't be easily optimized away).

Instead, given that `consume` already exists in 0.2 and isn't going anywhere, I'd suggest we make the minimal addition in 0.2.1 and just add a `take-headers`, and then consider our next major refactoring of the design as part of 0.3.

acfoltzer: That seems reasonable. To make the situation more generic, what we're really after are ways to take ownership of child resources in order to avoid deep-copying what those resources represent if they are needed elsewhere. Through that lens, the `variant`s and `string`s are indeed superfluous.

I think `take-headers` suffices for now since this PR is only augmenting the `incoming-` types which cannot be sent anywhere. It might be worth thinking about the behavior of such a method once incoming/outgoing are unified, e.g. what does it mean to send a request whose headers have been taken? There are a number of reasonable behaviors we might pick, such as the headers starting over from empty after being taken, or returning an error if trying a send-ater-take. A single static method that returns all of the child resources could be appealing to resolve that ambiguity.

wit/types.wit
acfoltzer: Removed these notes in 55a4754

wit/types.wit
lukewagner: Thanks!

lukewagner (approved):
Looks good to me.  It'd be good to hear if this makes sense to others too.

Assuming it does, before merging, we'll need to answer the process question of how we want to manage post-0.2 additions in the WIT in the individual proposal repos post-0.2.  I forget what we settled on.  In particular, once the contents of the proposal repo no longer match what's in the 0.2 snapshot in the `wasi` repo, should we first change all the versions in all the WITs in this repo to something other than `0.2.0`, say, `0.2.1-ephemeral`, so that we're not falsely claiming that our `take-headers` is part of 0.2.0?

Any thoughts @sunfishcode @ricochet @pchickey?

pchickey:
I believe we should change the version to be `0.2.1-draft` as part of this PR.

acfoltzer:
@pchickey I'm guessing we'll want to stick with 0.2.0 for the dependencies for now, and just change to `package wasi:http@0.2.1-draft` in `proxy.wit`?

pchickey:
Yes, we should only change the version at the `wasi:http` package declaration in `proxy.wit`.

acfoltzer:
Got it; updated the version and also edited the PR title to better reflect the new approach

acfoltzer:
Unsure why the CI check is failing here; I have `wit-bindgen-cli` 0.17.0 installed as does the Actions script, but the `--check` test is failing on the output I've produced.

PR #104, opened 2024-02-21
Add `http-response-status-invalid` error code

acfoltzer:
I failed to notice this was missing in #52. Status codes outside the range of 100-599 are [invalid](https://www.rfc-editor.org/rfc/rfc9110#section-15-6), though there is a nod to implementation-defined behavior outside those valid ranges. In practice we have found it very useful to return error codes specific to status problems to customers since our platform does not support codes outside this range, and the next-best option in the RFC 9209 equivalents is the generic `http-protocol-error` which doesn't help a user much with debugging.


pchickey:
We may not be able to accept this change before 0.3.0 - AFAIK, we are scaling back all of our subtyping aspirations to where just adding new functions to interfaces is permitted, and not changing any of the types defined in those interfaces.

acfoltzer:
Aha, that makes sense. I'm fine with putting this on ice then or changing the base revision to #101 if that's the best way to start accumulating breaking changes?

pchickey:
I guess in this repo we'll be working on 0.3.0-draft in parallel with 0.2.x-draft, so maybe landing #101 underneath a subdir (`0.3.0-draft/`?) and then we can keep iterating on both in `main`? @dicej @lukewagner wdyt?

lukewagner:
I like the idea of subdirs with versions as names; this matches what I expect we'll be doing as we release new official minor versions in the main `wasi` GH repo.

PR #106, opened 2024-03-07
add draft of wasi:http@0.3.0

dicej:
This adds a `wit-0.3.0-draft` directory so we can iterate on the 0.3.0 design in parallel with the 0.2.x work.  It's currently an exact copy of the `wit` directory except for the `handler.wit`, `proxy.wit`, and `types.wit` files, which have been updated to assume Component Model async, `stream`s, and `future`s are available.

High level description:

- The incoming-handler and outgoing-handler interfaces have been combined into a single handler interface.
- The incoming- and outgoing- variations of request, response, and body have been combined.
- I've added a option<request-options> field to request since it would be awkward to leave it as a parameter of handler.handle (e.g. what would it mean to receive such a parameter for an incoming request or for a request passed from one component to the other without any use of the network?).
- I've added a request-options-error (analogous to header-error) to distinguish between unsupported fields and immutable handles.
- We use `stream<u8>` and `future<trailers>` where appropriate.

You can find a working implementation of this API, plus several examples, in the [isyswasfa](https://github.com/dicej/isyswasfa) repository.

This PR supersedes #101.

### DIFF
--- a/imports.md
+++ b/imports.md
@@ -810,11 +810,9 @@ http-related errors.</p>
 which have multiple values are represented by multiple entries in this
 list with the same key.</p>
 <p>The tuple is a pair of the field key, represented as a string, and
-Value, represented as a list of bytes. In a valid Fields, all keys
-and values are valid UTF-8 strings. However, values are not always
-well-formed, so they are represented as a raw list of bytes.</p>
-<p>An error result will be returned if any header or value was
-syntactically invalid, or if a header was forbidden.</p>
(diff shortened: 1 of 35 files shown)


lukewagner (approved):
LGTM, nice work on all this!  I think anyone who wants to comment on the draft has had plenty of chance in #101, so I'll just wait a day before merging, in case anyone has any basic directory-layout or procedural comments.

PR #108, opened 2024-03-13
[0.3.0-draft] add `body.empty` static function

dicej:
This is useful for creating empty bodies, e.g. when sending GET requests. It avoids the extra ceremony of creating a stream and immediately dropping the writable handle.

### DIFF
--- a/wit-0.3.0-draft/types.wit
+++ b/wit-0.3.0-draft/types.wit
@@ -240,6 +240,9 @@ interface types {
       trailers: option<future<trailers>>
     );
 
+    /// Construct a new `body` with an empty stream and no trailers.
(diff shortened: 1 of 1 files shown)


tschneidereit:
Would it perhaps make sense to instead change the `request` and `response` ctor's `body` parameter to an `option<body>`? It seems unfortunate to incur the overhead of creating a resource and all that if it's never used.

dicej:
Would that mean we'd need to change e.g. `request.body` and `request.into-parts` to return `option<body>` as well?

tschneidereit:
I would think so, yes. And that seems to make sense to me, too

PR #110, opened 2024-03-15
[0.3.0-draft] return previous values from `fields.delete`

dicej:
This saves the user from having to use `get` followed by `delete` if they want access to the values they're removing.


lukewagner:
Since the CABI will force the copy to happen even if the value is immediately discarded, I wonder if we should have a separate `delete` and `get-and-delete`, with only the latter returning the just-deleted value?

dicej:
Superseded by https://github.com/WebAssembly/wasi-http/pull/111

PR #111, opened 2024-03-18
[0.3.0-draft] add `fields.get-and-delete`

dicej:
This saves the user from having to use `get` followed by `delete` if they want access to the values they're removing.


dicej:
Oops, modified the wrong file.  Will fix...

lukewagner (approved):

PR #115, opened 2024-05-30
Change `wasi:clocks` from `include` to `import` in `wasi:http/proxy`

yoshuawuyts:
This changes the `wasi:clocks` `include` statement to individual `import` statements in `wasi:http/proxy`. This came out of a conversation with @sunfishcode yesterday, who pointed out that if we're going to add timezones to `wasi:clocks` (https://github.com/WebAssembly/wasi-clocks/pull/61), `wasi:http/proxy` may not actually want to provide timezone support.

By switching from a catch-all `include` to individual `import` statements, we can decide later whether this is something we'd like to include. But conservatively we can start with the assumption that we won't.

This has some precedent in how we're handling our re-exports of `wasi:cli`: `wasi:http/proxy` includes `wasi:cli/stdin` and `wasi:cli/stdout` - but it does for example not include support for files or sockets.

### DIFF
--- a/imports.md
+++ b/imports.md
@@ -4,19 +4,172 @@ It is intended to be <code>include</code>d in other worlds.</p>
 <ul>
 <li>Imports:
 <ul>
+<li>interface <a href="#wasi_io_poll_0_2_0"><code>wasi:io/poll@0.2.0</code></a></li>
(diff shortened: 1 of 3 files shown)


lukewagner:
Makes sense to me

PR #122, opened 2024-07-25
Update request authority documentation

lann:
The existing wording refers to a "request's authority", which is not clearly defined by HTTP specs.

This rewords it to refer to the "request's target URI authority" - the "authority" part of the "target URI" of the request, which is well-defined: https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html#target.resource

Also normalize some word casing to match the HTTP Semantics doc.

### DIFF
--- a/imports.md
+++ b/imports.md
@@ -994,7 +994,7 @@ list with the same key.</p>
 <li><a name="method_incoming_request_scheme.0"></a> option&lt;<a href="#scheme"><a href="#scheme"><code>scheme</code></a></a>&gt;</li>
 </ul>
 <h4><a name="method_incoming_request_authority"></a><code>[method]incoming-request.authority: func</code></h4>
-<p>Returns the authority from the request, if it was present.</p>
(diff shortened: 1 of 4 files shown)


lann:
I'll fix up the branch if the rewording looks good to folks.

lukewagner (approved):
Ah, that's a nice succinct fix.  So just to check my understanding, by saying "authority of the target URI", you're implicitly using the "maybe get it from the `Host` header, maybe from the `:authority` pseudo-header, ..." spec-logic in https://www.rfc-editor.org/rfc/rfc9110#section-7.2, right?

Mossaka (approved):

lann:
The HTTP/2 spec is even more clear: https://www.rfc-editor.org/rfc/rfc7540.html#section-8.1.2.3

> The ":authority" pseudo-header field includes the authority portion of the target URI

lann:
Updated 0.3 draft and regenerated markdown.

ricochet (approved):

Issue #123, opened 2024-07-31
Can it be assumed that field keys have whitespace trimmed?


guybedford:
Field keys are currently specified as strings, and while validation is left to the implementation, it's not clear if implementations need to perform the trimming step still or not.

In a brief discussion with @lukewagner it sounded like this might be a possibility, and if so, it would be good to clarify this property.

guybedford:
I've updated the description here to separately consider keys and values.

lukewagner:
Good point, I think we do need to tighten this up in the prose wording.  I think the appropriate rule is that any `field-key` parameter (which we need to rename to `field-name`: #107) in wasi-http needs to be validated to match the [`field-name`](https://www.rfc-editor.org/rfc/rfc9110.html#name-field-names) grammar in RFC 9110 (which ends up disallowing whitespace, to your point) and, similarly, any `field-value` parameter in wasi-http must match the [`field-value`](https://www.rfc-editor.org/rfc/rfc9110.html#name-field-values) grammar in RFC 9110 (which ends up requiring the trimming), with a failure to match returning `invalid-syntax`.

PR #125, opened 2024-08-01
http: note field keys returned have whitespace trimmed

guybedford:
Resolves https://github.com/WebAssembly/wasi-http/issues/123, in allowing better optimizations for implementations since this is the only normalization (as opposed to validation) step required for field keys in the fetch specification.


guybedford:
Whitespace trimming applies to values not keys. Closing this one and updated the discussion in https://github.com/WebAssembly/wasi-http/issues/123 with further nuances.

PR #121, opened 2024-07-16
Specify the casing of field keys and values.

rylev:
Closes https://github.com/WebAssembly/wasi-http/issues/117

Issue #117, opened 2024-06-26
Perhaps be more specific about casing of `field-key`s


rylev:
[HTTP field keys are case *insensitive*](https://www.rfc-editor.org/rfc/rfc9110.html#section-5.1). However, `wasi-http` doesn't seem to specify that this should be the case when interacting with `fields`.

For example, `fields#get` likely should return values no matter the casing of the `field-key` passed in. This is how wasmtime's implementation currently works, but it doesn't seem specified anywhere that this *must* be the case. What's more, in wasmtime `fields#entries` returns tuples of the `field-key`/`field-value` pair where the `field-key` has the original casing. Is this the correct behavior or should hosts perhaps be expected to canonicalize these fields to lowercase instead?

pchickey:
Good point. @guybedford what does node do for this behavior?

guybedford:
Node.js automatically lowercases all header keys by the time they are provided to the user.

lukewagner:
Yes, good point, we need to improve the specified behavior here.  I'd suggest the following:
* Specify that field-name equality (as used by `fields.get`, `fields.set`, etc) is case insensitive (as defined by RFC 9110)
* The `fields` resource remembers the original case of field names (exactly as passed to `set` or as came in over the wire), and this remembered-case is what goes out over the wire.  This is what the JS Fetch API specifies because it was found necessary in practice for various interop scenarios.
* Unlike with the JS Fetch API's `Headers.get()`, which returns always-lowered, and `Headers.entries()`, which returns always-sorted, specify that WASI HTTP's `fields.get` and `fields.entries` returns the original case in the over-the-wire order so that it's possible for content to know exactly what will go out over the wire.

PR #127, opened 2024-08-14
Deprecate `field-key` in favor of `field-name`

### LINKED ISSUE #107: Terminology: "field key" vs "field name"
### Preface

Hi hi, I've been working with `wasi:http` a fair bit recently and have been enjoying it a lot. Thank you for that! Given we're currently still in a preview and we intend to make breaking changes, I'm going to interpret that as: *"Now is the time to give feedback on APIs, however small, because we can still make changes if we want to."*

### "field-key" vs "field-name"

Something small which stood out to me was the usage of the name [`field-key`](https://github.com/WebAssembly/wasi-http/blob/7f065d1d08e353aa105f02c2472b067d26e75e5c/wit/types.wit#L128-L129) to refer to the name of a field. I was expecting this to be called `field-name` instead. The reason for this is that in [RFC 9112: HTTP Semantics (2023)](https://httpwg.org/specs/rfc9112.html#header.field.syntax), each field is described as being composed of a `field-name` and a `field-value`:

```bnf
field-line = field-name ":" OWS field-value OWS
```
There is no mention of `field-key` anywhere in the RFC. The [IANA Field Name Registry](https://www.iana.org/assignments/http-fields/http-fields.xhtml) also refers to them as "fields", not "keys". I would like to propose we consider changing our terminology from "field-key" to "field-name".

### Navigating deprecations within a WASI release family

If we assume for a second this is a change we want to make, that opens up the question of how we would want to navigate a change like this. WASI 0.2 was just ratified, and WASI 0.3 is likely still a while out. And if we were to implement for example https://github.com/WebAssembly/wasi-http/issues/94 (add a `fields.keys` method), we'd be doubling-down on the existing terminology instead of the terminology we would prefer.

This is a question which is generally broader than this issue, but I think it would be good if we could establish a process for making changes like these *within* a minor WASI version. An example flow I'm thinking of is:

1. Introduce a new alias `type field-name = field-key` (or alias it directly to `string` if that's legal in our projections).
2. Change all APIs which take `field-key` to take a `field-name` instead.
3. Mark `field-key` as deprecated for WASI 0.2.
4. When we release WASI 0.3 we remove `field-key` from the API surface.

If we were able to do something like this we could release the changes as part of a WASI `0.2.x` release without breaking any existing code. And then in a WASI `0.3.0` release we could finalize the changes by actually removing deprecated APIs. I don't know whether we already have a way to support deprecating APIs in WIT; but if we don't that may be a good thing to consider adding.

yoshuawuyts:
Closes #107.

This deprecates `field-key` in favor of `field-name` using the `@deprecated` attribute in WIT. This is the first deprecation we're running in WASI, making this the first field test of it in our tooling. The way I've implemented the type change is by doing the following:

```wit
type field-name = field-key;
type field-key = string;
```

I don't know WIT's type system well enough to know whether I could instead just have written:

```wit
type field-name = string;
```

This PR also catches a case for #117 that #121 missed: only the `0.3.0-draft` APIs had their prose text updated about casing. That seemed to be on accident - and since we're touching the `field-key` API's text anyway, that seemed like something easy to include here.

Thanks!

### DIFF
--- a/wit-0.3.0-draft/types.wit
+++ b/wit-0.3.0-draft/types.wit
@@ -134,7 +134,7 @@ interface types {
 
   /// Field names are always strings.
   ///
-  /// Field keys should always be treated as case insensitive by the `fields`
+  /// Field names should always be treated as case insensitive by the `fields`
   /// resource for the purposes of equality checking.
   type field-name = string;
 
@@ -168,12 +168,12 @@ interface types {
     ///
     /// The resulting `fields` is mutable.
     ///
-    /// The list represents each key-value pair in the Fields. Keys
+    /// The list represents each name-value pair in the Fields. Names
     /// which have multiple values are represented by multiple entries in this
-    /// list with the same key.
+    /// list with the same name.
     ///
-    /// The tuple is a pair of the field key, represented as a string, and
-    /// Value, represented as a list of bytes. In a valid Fields, all keys
+    /// The tuple is a pair of the field name, represented as a string, and
+    /// Value, represented as a list of bytes. In a valid Fields, all names
     /// and values are valid UTF-8 strings. However, values are not always
     /// well-formed, so they are represented as a raw list of bytes.
     ///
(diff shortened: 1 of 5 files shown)


yoshuawuyts:
Changing this to a draft PR: `wit-abi-up-to-date` needs to be updated to the latest version of `wasm-tools` first. Once we've done that we can re-run CI.

yoshuawuyts:
This needs https://github.com/WebAssembly/wit-abi-up-to-date/pull/28 to land and be published so we can update `wit-tools` to a newer version, fixing CI.

wit-0.3.0-draft/types.wit
```diff
@@ -132,11 +132,11 @@ interface types {
     immutable,
   }
 
-  /// Field keys are always strings.
+  /// Field names are always strings.
   ///
   /// Field keys should always be treated as case insensitive by the `fields`
```
lukewagner, line 137:
```suggestion
  /// Field names should always be treated as case insensitive by the `fields`
```

lukewagner: Maybe do another grep -i for "key"; I see a few other "key"s hanging out in comments.

wit-0.3.0-draft/types.wit
```diff
@@ -154,8 +154,7 @@ interface types {
   /// `set`, `append`, and `delete` operations will fail with
   /// `header-error.immutable`.
   ///
-  /// A `fields` resource should store `field-key`s and `field-value`s in their
-  /// original casing used to construct or mutate the `fields` resource. The `fields`
+  /// A `fields` resource should store `field-name`s and `field-value`s in their /// original casing used to construct or mutate the `fields` resource. The `fields`
```
lukewagner, line 157:
```suggestion
  /// A `fields` resource should store `field-name`s and `field-value`s in their
  /// original casing used to construct or mutate the `fields` resource. The `fields`
```

lukewagner (approved):
Thanks!  Cool to see @deprecated in action.

yoshuawuyts:
Realized just now that we'd set the `@deprecated` gate to point to version `0.2.1` which has already shipped, rather than `0.2.2` which is the next release. It seems like we missed that in review.

`wit-abi-up-to-date` has been updated to allow the `wasm-tools` version to be set as a parameter. Which means CI is now passing for this PR. Going to go ahead and merge this! Thanks all!

PR #143, opened 2025-01-17
feat(0.3): rework stream error handling/remove `wasi:io` usage

rvolosatovs:
Remove `wasi:io` usage in `wit-0.3.0-draft`

Followed the same stream error handling convention as currently in
- https://github.com/WebAssembly/wasi-filesystem/pull/164
- https://github.com/WebAssembly/wasi-sockets/pull/111

### DIFF
--- a/wit-0.3.0-draft/types.wit
+++ b/wit-0.3.0-draft/types.wit
@@ -2,7 +2,6 @@
 /// Requests and Responses, as well as their headers, trailers, and bodies.
 interface types {
   use wasi:clocks/monotonic-clock@0.3.0.{duration};
-  use wasi:io/error@0.2.3.{error};
 
   /// This type corresponds to HTTP standard Methods.
   variant method {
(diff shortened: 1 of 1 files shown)


lukewagner (approved):
The change to the `constructor`/`new` is actually pretty interesting, because it demonstrates how the "return a `future<result<_, error-code>>`" trick is necessary not just for functions *returning* a `stream`, but also functions *taking* a `stream`, and thus even the `stream<T,SuccessT,ErrorT>` extension (presented [here](https://docs.google.com/presentation/d/1Is0yTZ13xuw5UjbbUH_h33HVqVEjDV3XDzQFWmUHolg/edit#slide=id.g327f93fb0cd_0_6)) wouldn't eliminate it.  This raises some interesting questions on whether the `ErrorT` should be able to flow both ways... but that's all for later, and I think what you're doing here is about the only possible thing we can do in a 0.3.0 timeframe to not regress from 0.2, which provided `error-code`s upon write failures.  Nice work!  (cc @dicej )

rvolosatovs:
@lukewagner I've rebased on latest `main` to pull in the change from #142 and resolve a merge conflict

lukewagner (approved):
Given all the things in flight, I wasn't sure when's a good point to merge (and iterate in later PRs), but seems fine to merge now?

rvolosatovs:
> Given all the things in flight, I wasn't sure when's a good point to merge (and iterate in later PRs), but seems fine to merge now?

I think it's fine to merge now indeed

PR #152, opened 2025-02-07
Fix CI by enabling all features

yoshuawuyts:
This is an alternative solution to https://github.com/WebAssembly/wasi-http/pull/151 using https://github.com/WebAssembly/wit-abi-up-to-date/pull/29 to enable all features in CI. This approach has precedent in e.g. `wasi:cli`, where we [generate docs](https://github.com/WebAssembly/wasi-cli/blob/main/command.md#exit-with-code-func) for the unstable `cli-exit-with-code` feature.

Assuming this PR passes CI, #151 can be trimmed down to just upgrading the 0.3 draft's dependencies. And in a future PR we should address the issue @rvolosatovs pointed out where PRs filed by GitHub Actions do not trigger CI runs.

Thanks!


wit-0.3.0-draft/deps.lock
```diff
@@ -1,12 +1,12 @@
 [cli]
-url = "https://github.com/WebAssembly/wasi-cli/archive/v0.2.3.tar.gz"
```
ricochet, line 2: 0.3-draft in the lock but this does fix ci for now

ricochet (approved):

PR #154, opened 2025-02-19
feat: allow independent retrieval of trailers

vados-cosmonic:
This commit changes wasi:http/types#body.finish to forward the optionally present trailers, rather than retrieve them with the possibility of generating an error along the way.

With this change, callers of `body.finish` will have to retreive trailers on their own, which *may* produce an efficiency gain in the case where trailers are present but not needed, and avoiding work of checking for trailers in the first place.

### :eyes: Implementation preview

As this is a WASI p3 feature (for which work is ongoing in the [wasip3-prototyping repository](https://github.com/bytecodealliance/wasip3-prototyping)), you can "preview" the effects of this change on the implementation [in the PR there](https://github.com/bytecodealliance/wasip3-prototyping/pull/16).

### DIFF
--- a/wit-0.3.0-draft/types.wit
+++ b/wit-0.3.0-draft/types.wit
@@ -240,14 +240,27 @@ interface types {
   /// trailers at any given time.
   resource body {
 
-    /// Construct a new `body` with the specified stream and trailers.
+    /// Construct a new `body` with the specified stream.
+    ///
     /// This function returns a future, which will resolve
-    /// to an error code if transmitting stream data or trailers fails.
-    /// The returned future resolves to success once body stream and trailers
-    /// are fully transmitted.
+    /// to an error code if transmitting stream data fails.
+    ///
+    /// The returned future resolves to success once body stream
+    /// is fully transmitted.
     new: static func(
       %stream: stream<u8>,
-      trailers: option<future<trailers>>
+    ) -> tuple<body, future<result<_, error-code>>>;
+
+    /// Construct a new `body` with the specified stream and trailers.
+    ///
+    /// This function returns a future, which will resolve
+    /// to an error code if transmitting stream data or trailers fails.
+    ///
+    /// The returned future resolves to success once body stream and trailers
+    /// are fully transmitted.
+    new-with-trailers: static func(
+      %stream: stream<u8>,
+      trailers: future<trailers>
     ) -> tuple<body, future<result<_, error-code>>>;
 
     /// Returns the contents of the body, as a stream of bytes.
@@ -260,9 +273,10 @@ interface types {
     /// The returned future resolves to success if body is closed.
     %stream: func() -> result<tuple<stream<u8>>, future<result<_, error-code>>>;
 
-    /// Takes ownership of `body`, and returns a `trailers`.  This function will
-    /// trap if a `stream` child is still alive.
-    finish: static func(this: body) -> result<option<trailers>, error-code>;
+    /// Takes ownership of `body`, and returns an unresolved optional `trailers`.
+    ///
+    /// This function will trap if a `stream` child is still alive.
+    finish: static func(this: body) -> future<option<trailers>>;
   }
 
   /// Represents an HTTP Request.


wit-0.3.0-draft/types.wit
```diff
@@ -260,9 +260,10 @@ interface types {
     /// The returned future resolves to success if body is closed.
     %stream: func() -> result<tuple<stream<u8>>, future<result<_, error-code>>>;
 
-    /// Takes ownership of `body`, and returns a `trailers`.  This function will
-    /// trap if a `stream` child is still alive.
-    finish: static func(this: body) -> result<option<trailers>, error-code>;
+    /// Takes ownership of `body`, and returns an unresolved optional `trailers`.
+    ///
+    /// This function will trap if a `stream` child is still alive.
+    finish: static func(this: body) -> option<future<trailers>>;
```
pchickey, line 266:
```suggestion
    finish: static func(this: body) -> future<option<trailers>>;
```

dicej: Yes, I agree with this suggestion.  The implementation of `finish` might not know whether trailers are on their way at the time when `finish` is called, and it should not wait to find out while the caller is waiting for a result.  If the caller cares about trailers at all (and most won't), it can await the future to discover whether there are any and, if so, what they are.

vados-cosmonic: Yeah this looks this way as it's what the `Body` gives us -- will need to work out a way to push the `Option` into the `FutureReader`, now that we have removed the unnecessary `Result`

vados-cosmonic: Yeah this change has to go into into `body.new` as well -- so will make that change as well and see what the interface looks like then

vados-cosmonic: OK, updated and push the change through to the [impl PR](https://github.com/bytecodealliance/wasip3-prototyping/pull/16) as well

dicej:
@vados-cosmonic I'm wondering if we should stick with the `option<future<trailers>>` param type in the `body` constructor.  I realize it's asymmetric with respect to `body.finish`, but it's arguably ideal from an ergonomics standpoint given that most applications will neither send nor receive trailers.  When creating a new body without trailers, the easiest thing to do is specify `none` rather than go through the motions of creating a new future just to send `none` to it (or close it without sending anything, which I suppose would be morally equivalent).

Obviously, the asymmetry implies that the implementation will need to transpose a `option<future<trailers>>` to a `future<option<trailers>>` in some scenarios, but I expect that won't be difficult.

dicej:
Not to complicate things further, but we could even make `body.finish` return a `future<trailers>` and assume the write end will close the future without sending anything in the case there are no trailers.

vados-cosmonic:
Hey @dicej waht do you think about actually changing the constructor for body to one without trailers and separating the ability to add/set trailers (either a different constructor or a setter are both fine).

> Obviously, the asymmetry implies that the implementation will need to transpose a option<future<trailers>> to a future<option<trailers>> in some scenarios, but I expect that won't be difficult.

So I found this difficult to do in the impl without creating another future during `finish` -- transposing the `Option` into the `FutureReader` goes across the async boundary and the types don't quite permit me to change the `T`to an `Option<T>` whether it's in the host machinery or otherwise. Maybe I was missing something obvious!

> Not to complicate things further, but we could even make body.finish return a future<trailers> and assume the write end will close the future without sending anything in the case there are no trailers.

This is certainly an interesting idea -- does this make it a little more difficult in that the write end has to remember to close the future? If this introduces the possibility of a hang due to incorrectly implemented writers it might increase the hurdle.

This also kind of makes me want to separate the common case and the with-trailers case into different functions, what do you think?

dicej:
> So I found this difficult to do in the impl without creating another future during `finish` -- transposing the `Option` into the `FutureReader` goes across the async boundary and the types don't quite permit me to change the `T`to an `Option<T>` whether it's in the host machinery or otherwise. Maybe I was missing something obvious!

Yeah, you'd need to create another future, although I think we can avoid that if `finish` returns `future<trailers>` instead of `future<option<trailers>>`.

> This is certainly an interesting idea -- does this make it a little more difficult in that the write end has to remember to close the future?

They will either need to remember to send `none` or close the future, and I don't think the latter is more onerous than the former.

dicej:
BTW, I'm also fine with removing the trailers parameter from the constructor and e.g. adding a new `new-with-trailers: static func(%stream: stream<u8>, trailers: future<trailers>) -> body;` function.

vados-cosmonic:
> Yeah, you'd need to create another future, although I think we can avoid that if finish returns future<trailers> instead of future<option<trailers>>.

I'm in favor of this!

> BTW, I'm also fine with removing the trailers parameter from the constructor and e.g. adding a new new-with-trailers: static func(%stream: stream<u8>, trailers: future<trailers>) -> body; function.

A *tiny* nit, but I personally like just `with-trailers`. 

Will re-write the impl and make updates here with that in mind.

wit-0.3.0-draft/types.wit
vados-cosmonic: Going with the discussion below (outside this comment), I've update the impl and the WIT yet again. 

The prototyping repo's WITs aren't *quite* up to date, but the changes here are structurally reflected there (new constructor & changed `finish` signature)

dicej (approved):

pchickey (approved):

dicej:
Okay, turns out I was wrong about how `future` works.  The owner of the write end of a future is in fact _not_ allowed to close it without either writing a value to it or closing it with an `error-context`.  So my suggestion that we could use `future<trailers>` and interpret a close-without-write as "there were no trailers" rather than "an error occurred" is invalid.

That means that we must make `body.finish` return either `option<future<option<trailers>>` or `future<option<trailers>>` if there's a chance that the implementation won't know for sure whether trailers are on their way by the time it returns a result.  I.e. returning `option<future<trailers>>` would imply that implementations have read to the end of the content stream before returning, which I don't think is what we want (or if it _is_ what we want, we can just go back to returning `option<trailers>` and assume `finish` will be called async).  Does that make sense?

PR #159, opened 2025-03-12
fix(p3): allow taking response/request body at most once

rvolosatovs:
Similar to #156, specify that the request/response body can be taken only once.

If callers were able to get N references to the body, then they would be able to call `body.finish` N times, which would require multiplexing in the host to provide N copies of e.g. the trailers to these N references. It's hard to imagine a use case for this pattern and it also not a common pattern in HTTP libraries as far as I know, so specify that body can only be taken once.

### DIFF
--- a/wit-0.3.0-draft/types.wit
+++ b/wit-0.3.0-draft/types.wit
@@ -356,6 +356,8 @@ interface types {
 
     /// Get the body associated with the Request, if any.
     ///
+    /// After being called once, consequent calls with return `none`.
+    ///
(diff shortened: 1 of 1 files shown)


lukewagner:
Similar to my reply in #156, the idea is that a request/response can be handed to component A, it can call `body` to process a part of the body, and then drop the body handle, and then pass the request/response to component B, which will call `body` again, picking up where component A left off.  That would suggest we add some verbiage to the same effect as on `%stream` where "you can only call this a second time after dropping the original body".

PR #160, opened 2025-03-12
feat(p3): return owned `request-options` on consumption

rvolosatovs:
Currently, there's no way to acquire the owned `request-options` from the `request`, return them as part of the `into-parts` along the other moved resources

### DIFF
--- a/wit-0.3.0-draft/types.wit
+++ b/wit-0.3.0-draft/types.wit
@@ -361,9 +361,9 @@ interface types {
     /// component by e.g. `handler.handle`.
     body: func() -> option<body>;
 
-    /// Takes ownership of the `request` and returns the `headers` and `body`,
-    /// if any.
-    into-parts: static func(this: request) -> tuple<headers, option<body>>;
+    /// Takes ownership of the `request` and returns the `headers`, `body`
+    /// and `request-options`, if any.
+    into-parts: static func(this: request) -> tuple<headers, option<body>, option<request-options>>;
   }
 
   /// Parameters for making an HTTP Request. Each of these parameters is


lukewagner:
Makes sense to me

dicej (approved):

dicej:
I've already merged this, but I just had one more thought.  What if we did something like this:

```fsharp
record request-parts {
    method: method,
    path_with_query: option<string>,
    scheme: option<scheme>,
    authority: option<string>,
    headers: headers,
    options: option<request-options>,
    body: option<body>,
}

resource request {
    // ...

    into-parts: static func(this: request) -> request-parts;
    from-parts: static func(parts: request-parts) -> request;
}
```

(and likewise for `response`/`response-parts`)

That could make request/response-modifying middleware easier to write -- just call `into-parts`, tweak the fields you want to, and then call `from-parts`.

Thoughts?

rvolosatovs:
> I've already merged this, but I just had one more thought.  What if we did something like this:
> 
> ```fsharp
> record request-parts {
>     method: method,
>     path_with_query: option<string>,
>     scheme: option<scheme>,
>     authority: option<string>,
>     headers: headers,
>     options: option<request-options>,
>     body: option<body>,
> }
> 
> resource request {
>     // ...
> 
>     into-parts: static func(this: request) -> request-parts;
>     from-parts: static func(parts: request-parts) -> request;
> }
> ```
> 
> (and likewise for `response`/`response-parts`)
> 
> That could make request/response-modifying middleware easier to write -- just call `into-parts`, tweak the fields you want to, and then call `from-parts`.
> 
> Thoughts?

Assuming that the cost of moving the field values is negligible, I think that would work great - I was thinking about doing the same, but opted for the most minimal, incremental change I could come up with.

dicej:
If we're concerned about the overhead of lifting and lowering all the fields, we could do something like this:

```fsharp
record request-parts {
    request: request, // Request with no headers, options, or body (i.e. just method, path-with-query, etc.)
    headers: headers,
    options: option<request-options>,
    body: option<body>,
}

resource request {
    // ...

    into-parts: static func(this: request) -> request-parts;
    from-parts: static func(parts: request-parts) -> request;
}
```

PR #162, opened 2025-03-13
[v0.3] model bodies after `stream<u8, trailers, error-code>`, trap less

### LINKED ISSUE #163: [v0.3] `handle` implementations cannot forward trailers without trapping while handling transmit errors
Components exporting `wasi:http@0.3.0-draft/handle.handle` cannot correctly forward incoming requests while tracking the body transmit error.

Assume an application, which forwards the incoming request and returns a successful (200 OK) response with empty body **iff forwarding of the incoming request in it's entirety has succeeded to the best of it's knowledge** (we can never know for sure due to OS-level buffering).

Pseudo-code (in pseudo-Rust):
```rust
async fn handle(req: Request) -> Result<Response, ErrorCode> {
    let headers = req.headers().clone();
    headers.set("X-Forwarded-For", origin_of(headers));

    let req_body = req.body()?;
    let (req_stream, req_result) = req_body.stream()?;
    let (trailers_tx, trailers_rx) = future::new();
    // We don't know if we will receive trailers at this point, so we are forced to use this API
    let (out_body, out_result) = body::new_with_trailers(req_stream, trailers_rx);

    // NOTE: this is incorrect, trailers might not be present and we'll only know about it once we have received the body
    // We should not send empty trailers if none were given, but closing the future without a result is not valid according to the spec

    let res = http::handle(Request::new(headers, Some(out_body), None)).await?;

    let trailers = body::finish(req_body).await;
    // TRAP!
}
```

- The application will only find out whether there are trailers to forward once the full incoming body has been read, however the body constructor forces us to make a choice before that has happened (since we want to *stream* the body being forwarded)
- The only way for the application to get a handle to the trailers is by calling `finish` on the incoming body, however we cannot do that, since there is an active `stream` handle, that we must pass to the `body` constructor.
   - Even if #162 was to be merged, this would still be broken - we'd need to only call `finish` once the body has been fully streamed, but we don't have a way of knowing when that has happened.

It seems that we need to change the trailers future signature to return `option<trailers>` and have some way to acquire trailers from the body without moving it

Refs #164

### LINKED ISSUE #164: [p3] `handle` implementations have no way of reporting errors occuring after response return
Assume the following pseudo-Rust pseudocode implementation of `handle`, which streams contents of a file represented by some descriptor `fd` using [`read-via-stream`](https://github.com/WebAssembly/wasi-filesystem/blob/740cd7617b52eb54cdf4f2f038cfc6a42f604c3f/wit-0.3.0-draft/types.wit#L305-L308) in response to an incoming request

```
async fn handle(req: request) -> result<response, error_code> {
    let (fd_data, fd_result) = fd.read_via_stream(0).await?;
    let (body, tx_result) = body::new(fd_data, None);
    spawn({
        if fd_result.await.is_err() {
            // where to report result?
        }
    });
    return Ok(response.new(headers.new(), Some(body)));
}
```

An I/O error might have occurred reading from `fd` *after* `handle` has returned, however there is no way for the application to report that error to the host other that somehow using the "application layer", i.e. HTTP body or trailers, to communicate it.

rvolosatovs:
- Make bodies non-optional for request and response constructors
- Take a `future<result<option<trailers>, error-code>>` parameter in request/response constructors (refs https://github.com/WebAssembly/component-model/issues/474#issuecomment-2736430031)
- Report any transmit errors occurring in the future returned by request/response constructor
- Allow `headers` resources acquired by calling `request.headers` or `response.headers` to persist *after* respective `into-parts` is called, referencing the immutable headers of the original request/response. This implies that the `headers` acquired this way are not, if fact, children of the original request/response they originated from and the original request/response can be freely dropped and/or moved even before they are dropped.
- Do not trap if guest attempts to acquire multiple handles to the body stream - handle that using type system
- Model bodies after future `stream<u8, trailers, error-code>`
- Remove `into-parts`
- Add `request-options#clone`

Closes https://github.com/WebAssembly/wasi-http/pull/156
Closes https://github.com/WebAssembly/wasi-http/pull/161
Closes #164 
Closes #163 
Closes #165

### DIFF
--- a/wit-0.3.0-draft/types.wit
+++ b/wit-0.3.0-draft/types.wit
@@ -265,17 +265,26 @@ interface types {
 
     /// Returns the contents of the body, as a stream of bytes.
     ///
-    /// This function may be called multiple times as long as any `stream`s
-    /// returned by previous calls have been dropped first.
-    ///
-    /// On success, this function returns a stream and a future, which will resolve
+    /// This function returns a stream and a future, which will resolve
     /// to an error code if receiving data from stream fails.
-    /// The returned future resolves to success if body is closed.
-    %stream: func() -> result<tuple<stream<u8>, future<result<_, error-code>>>>;
+    /// The returned future resolves to success if data section of the
+    /// body is fully consumed.
+    ///
+    /// The handles returned by this function are considered to be child
+    /// handles. Dropping the resource on which this function is called
+    /// will close the handles with an error context, if they are still open.
+    ///
+    /// This function may be called multiple times.
+    /// If it is called while either a stream or future returned by a previous 
+    /// call to this function is still open, those handles will be closed 
+    /// with an error context.
+    %stream: func() -> tuple<stream<u8>, future<result<_, error-code>>>;
 
     /// Takes ownership of `body`, and returns an unresolved optional `trailers` result.
     ///
-    /// This function will trap if a `stream` child is still alive.
+    /// Developers are encouraged to close any open handles returned by previous
+    /// calls to `stream`, if any, before calling this function,
+    /// but calling `finish` will close them with an error context.
     finish: static func(this: body) -> future<result<option<trailers>, error-code>>;
   }
 
@@ -356,13 +365,30 @@ interface types {
 
     /// Get the body associated with the Request, if any.
     ///
-    /// This body resource is a child: it must be dropped before the parent
-    /// `request` is dropped, or its ownership is transferred to another
-    /// component by e.g. `handler.handle`.
+    /// This body resource is a child: it must be dropped or consumed before
+    /// the parent `request` is dropped, or its ownership is transferred
+    /// to another component by e.g. `handler.handle`.
+    ///
+    /// Only a single body resource can be active at a time.
+    /// If this function is called before the body resource returned by
+    /// a previous call to this function is dropped, the previously-returned
+    /// body will be transitioned into an "error state" and all operations
+    /// on it will fail.
+    ///
+    /// In case, that a `body.finish` was called on the body returned by a previous
+    /// call to this function, this function returns `none`.
     body: func() -> option<body>;
 
     /// Takes ownership of the `request` and returns the `headers`, `body`
     /// and `request-options`, if any.
+    ///
+    /// The headers returned by this function are considered to be a clone of the headers
+    /// of the request. Changes to the `headers` returned by this function will not be reflected
+    /// in the immutable `headers` resources, that could have been acquired by calls to `headers`
+    /// prior to this function being called.
+    ///
+    /// In case, that a `body.finish` was called on the body returned by a previous
+    /// call to `body`, this function returns body as `none`.
     into-parts: static func(this: request) -> tuple<headers, option<body>, option<request-options>>;
   }
 
@@ -439,13 +465,30 @@ interface types {
 
     /// Get the body associated with the Response, if any.
     ///
-    /// This body resource is a child: it must be dropped before the parent
-    /// `response` is dropped, or its ownership is transferred to another
-    /// component by e.g. `handler.handle`.
+    /// This body resource is a child: it should be dropped or consumed before
+    /// the parent `response` is dropped, or its ownership is transferred
+    /// to another component by e.g. `handler.handle`.
+    ///
+    /// Only a single body resource can be active at a time.
+    /// If this function is called before the body resource returned by
+    /// a previous call to this function is dropped, the previously-returned
+    /// body will be transitioned into an "error state" and all operations
+    /// on it will fail.
+    ///
+    /// In case, that a `body.finish` was called on the body returned by a previous
+    /// call to this function, this function returns `none`.
     body: func() -> option<body>;
 
     /// Takes ownership of the `response` and returns the `headers` and `body`,
     /// if any.
+    ///
+    /// The headers returned by this function are considered to be a clone of the headers
+    /// of the response. Changes to the `headers` returned by this function will not be reflected
+    /// in the immutable `headers` resources, that could have been acquired by calls to `headers`
+    /// prior to this function being called.
+    ///
+    /// In case, that a `body.finish` was called on the body returned by a previous
+    /// call to `body`, this function returns body as `none`.
     into-parts: static func(this: response) -> tuple<headers, option<body>>;
   }
 }
--- a/wit-0.3.0-draft/types.wit
+++ b/wit-0.3.0-draft/types.wit
@@ -265,29 +265,28 @@ interface types {
 
     /// Returns the contents of the body, as a stream of bytes.
     ///
-    /// This function returns a stream and a future, which will resolve
+    /// This method returns a stream and a future, which will resolve
     /// to an error code if receiving data from stream fails.
     /// The returned future resolves to success if data section of the
     /// body is fully consumed.
     ///
-    /// The handles returned by this function are considered to be child
-    /// handles. Dropping the resource on which this function is called
+    /// The handles returned by this method are considered to be child
+    /// handles. Dropping the resource on which this method is called
     /// will close the handles with an error context, if they are still open.
     ///
-    /// This function may be called multiple times.
-    /// If it is called while either a stream or future returned by a previous 
-    /// call to this function is still open, those handles will be closed 
-    /// with an error context. Thus there will always be at most one readable stream
-    /// open for a given body. Each subsequent stream picks up where the last
-    /// stream left off, up until `finish` is called.
-    %stream: func() -> tuple<stream<u8>, future<result<_, error-code>>>;
+    /// This method may be called multiple times.
+    /// If it is called while either a stream or future returned by a previous
+    /// call to this method is still open, this method will return an error.
+    /// Thus there will always be at most one readable stream open for a given body.
+    /// Each subsequent stream picks up where the last stream left off,
+    /// up until `finish` is called.
+    %stream: func() -> result<tuple<stream<u8>, future<result<_, error-code>>>>;
 
     /// Takes ownership of `body`, and returns an unresolved optional `trailers` result.
     ///
-    /// Developers are encouraged to close any open handles returned by previous
-    /// calls to `stream`, if any, before calling this function,
-    /// but calling `finish` will close them with an error context.
-    finish: static func(this: body) -> future<result<option<trailers>, error-code>>;
+    /// If this function is called while either a stream or future returned by a previous
+    /// call to `body.stream` is still open, this function will return an error.
+    finish: static func(this: body) -> result<future<result<option<trailers>, error-code>>>;
   }
 
   /// Represents an HTTP Request.
@@ -368,14 +367,12 @@ interface types {
     /// to another component by e.g. `handler.handle`.
     ///
     /// Only a single body resource can be active at a time.
-    /// If this function is called before the body resource returned by
-    /// a previous call to this function is dropped, the previously-returned
-    /// body will be transitioned into an "error state" and all operations
-    /// on it will fail.
+    /// If this method is called before the body resource returned by
+    /// a previous call to this method is dropped, error will be returned.
     ///
     /// Once `body.finish` is called on a body returned by this method,
-    /// all subsequent calls to this method will return `none`.
-    body: func() -> option<body>;
+    /// all subsequent calls to this method will return `ok(none)`.
+    body: func() -> result<option<body>>;
 
     /// Takes ownership of the `request` and returns the `headers`, `body`
     /// and `request-options`, if any.
@@ -384,6 +381,8 @@ interface types {
     /// of the request. Changes to the `headers` returned by this function will not be reflected
     /// in the immutable `headers` resources, that could have been acquired by calls to `headers`
     /// prior to this method being called.
+    ///
+    /// If a body child exists when this function is called, an error will be returned.
     ///
     /// Once `body.finish` is called on a body returned by a previous
     /// call to `body` method, this method returns body as `none`.
@@ -464,14 +463,12 @@ interface types {
     /// to another component by e.g. `handler.handle`.
     ///
     /// Only a single body resource can be active at a time.
-    /// If this function is called before the body resource returned by
-    /// a previous call to this function is dropped, the previously-returned
-    /// body will be transitioned into an "error state" and all operations
-    /// on it will fail.
+    /// If this method is called before the body resource returned by
+    /// a previous call to this method is dropped, error will be returned.
     ///
     /// Once `body.finish` is called on a body returned by this method,
-    /// all subsequent calls to this method will return `none`.
-    body: func() -> option<body>;
+    /// all subsequent calls to this method will return `ok(none)`.
+    body: func() -> result<option<body>>;
 
     /// Takes ownership of the `response` and returns the `headers` and `body`,
     /// if any.
@@ -481,8 +478,10 @@ interface types {
     /// in the immutable `headers` resources, that could have been acquired by calls to `headers`
     /// prior to this method being called.
     ///
+    /// If a body child exists when this function is called, an error will be returned.
+    ///
     /// Once `body.finish` is called on a body returned by a previous
     /// call to `body` method, this method returns body as `none`.
-    into-parts: static func(this: response) -> tuple<headers, option<body>>;
+    into-parts: static func(this: response) -> result<tuple<headers, option<body>>>;
   }
 }
--- a/wit-0.3.0-draft/types.wit
+++ b/wit-0.3.0-draft/types.wit
@@ -230,79 +230,35 @@ interface types {
   /// Trailers is an alias for Fields.
   type trailers = fields;
 
-  /// Represents an HTTP Request or Response's Body.
-  ///
-  /// A body has both its contents - a stream of bytes - and a (possibly empty)
-  /// set of trailers, indicating that the full contents of the body have been
-  /// received. This resource represents the contents as a `stream<u8>` and the
-  /// delivery of trailers as a `trailers`, and ensures that the user of this
-  /// interface may only be consuming either the body contents or waiting on
-  /// trailers at any given time.
-  resource body {
-
-    /// Construct a new `body` with the specified stream and `finish` future.
-    ///
-    /// The `finish` future must resolve to a result once `stream` has finished.
-    ///
-    /// This function returns a future, which will resolve to an error code if
-    /// transmitting of the request or response this body is a part of fails.
-    ///
-    /// The returned future resolves to success once request or response this
-    /// body is a part of is fully transmitted.
-    new: static func(
-      %stream: stream<u8>,
-      finish: future<result<option<trailers>, error-code>>,
-    ) -> tuple<body, future<result<_, error-code>>>;
-
-    /// Returns the contents of the body, as a stream of bytes.
-    ///
-    /// This method returns a stream and a future, which will resolve
-    /// to an error code if receiving data from stream fails.
-    /// The returned future resolves to success if data section of the
-    /// body is fully consumed.
-    ///
-    /// The handles returned by this method are considered to be child
-    /// handles. Dropping the resource on which this method is called
-    /// will close the handles with an error context, if they are still open.
-    ///
-    /// This method may be called multiple times.
-    /// This method will return an error if it is called while either:
-    /// - a stream or future returned by a previous call to this method is still open
-    /// - `finish` was called on another instance of this `body` resource
-    /// Thus there will always be at most one readable stream open for a given body.
-    /// Each subsequent stream picks up where the last stream left off,
-    /// up until `finish` is called on any of the instances of this `body` resource
-    %stream: func() -> result<tuple<stream<u8>, future<result<_, error-code>>>>;
-
-    /// Takes ownership of `body`, and returns an unresolved optional `trailers` result.
-    ///
-    /// This function will return an error if it is called while either:
-    /// - a stream or future returned by a previous call to `body.stream` is still open
-    /// - `finish` was already called on another instance of this `body` resource
-    finish: static func(this: body) -> result<future<result<option<trailers>, error-code>>>;
-  }
-
   /// Represents an HTTP Request.
   resource request {
 
     /// Construct a new `request` with a default `method` of `GET`, and
     /// `none` values for `path-with-query`, `scheme`, and `authority`.
     ///
-    /// * `headers` is the HTTP Headers for the Response.
-    /// * `body` is the contents of the body, possibly including trailers.
-    /// * `options` is optional `request-options` to be used if the request is
-    ///   sent over a network connection.
+    /// `headers` is the HTTP Headers for the Request.
+    ///
+    /// `contents` is the body content stream. Once it is closed,
+    /// `trailers` future must resolve to a result.
+    /// If `trailers` resolves to an error, underlying connection
+    /// will be closed immediately.
+    ///
+    /// `options` is optional `request-options` resource to be used
+    /// if the request is sent over a network connection.
     ///
     /// It is possible to construct, or manipulate with the accessor functions
-    /// below, an `request` with an invalid combination of `scheme`
+    /// below, a `request` with an invalid combination of `scheme`
     /// and `authority`, or `headers` which are not permitted to be sent.
     /// It is the obligation of the `handler.handle` implementation
     /// to reject invalid constructions of `request`.
-    constructor(
+    ///
+    /// The returned future resolves to result of transmission of this request.
+    new: static func(
       headers: headers,
-      body: body,
+      contents: stream<u8>,
+      trailers: future<result<option<trailers>, error-code>>,
       options: option<request-options>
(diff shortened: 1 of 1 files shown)


wit-0.3.0-draft/types.wit
```diff
@@ -348,21 +357,34 @@ interface types {
...
+    /// call to this function, this function returns `none`.
     body: func() -> option<body>;
 
     /// Takes ownership of the `request` and returns the `headers`, `body`
     /// and `request-options`, if any.
+    ///
+    /// The headers returned by this function are considered to be a clone of the headers
+    /// of the request. Changes to the `headers` returned by this function will not be reflected
+    /// in the immutable `headers` resources, that could have been acquired by calls to `headers`
+    /// prior to this function being called.
```
lukewagner, line 384: I know this is what we discussed, and it is the friendliest option but I remembered later that, when we discussed this as a group in the BA C-M impl call, the idea was to instead say: if you call `into-parts` when there is already an existing handle to `headers`, that existing `headers` handle is put into a state where all header operations fail (how precisely, to be defined per-method).  This is symmetric to what you're describing with `body` and it matches more-closely how a built-in WIT `child use<headers>` return type would work (bindgen wouldn't know to clone).  WDYT?

(and similarly in `response`)

wit-0.3.0-draft/types.wit
```diff
@@ -348,21 +357,34 @@ interface types {
...
+    /// to another component by e.g. `handler.handle`.
+    ///
+    /// Only a single body resource can be active at a time.
+    /// If this function is called before the body resource returned by
+    /// a previous call to this function is dropped, the previously-returned
+    /// body will be transitioned into an "error state" and all operations
+    /// on it will fail.
+    ///
+    /// In case, that a `body.finish` was called on the body returned by a previous
+    /// call to this function, this function returns `none`.
```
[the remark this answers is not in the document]
lukewagner: (and similarly in `response`)

lukewagner (commented):
Thanks for writing this up and thinking about lifetimes.  Overall, this makes sense as part of the general goal (that we discussed in the earlier BA C-M impl meeting) of, in cases where a child would otherwise outlive its parent, having lazier tombstone/failure semantics.

wit-0.3.0-draft/types.wit
rvolosatovs: It seems like we'd need to wrap `fields#get`, `fields#has` and `fields#entries` results in a `result<T, header-error>`, as well as introduce a `header-error#moved` or `header-error#corrupted` to fully cover this case.
- https://github.com/WebAssembly/wasi-http/blob/9caa8b30d615bc9583ccf3d346783ebae8a33e4a/wit-0.3.0-draft/types.wit#L174-L182
- https://github.com/WebAssembly/wasi-http/blob/9caa8b30d615bc9583ccf3d346783ebae8a33e4a/wit-0.3.0-draft/types.wit#L210-L219

Otherwise, I'm afraid this behavior would inevitably cause bugs in applications using this API.

I feel like the behavior proposed in this PR is less surprising, provide better developer experience and introduce no overhead. That said I'm happy to add a new `header-error` and wrap the methods with a result, if you think we should do that. Otherwise, perhaps we should discuss this in next BA C-M impl call?

tschneidereit:
> * Auto-close any existing, active body resources on repeated calls to `request.body` or `response.body`
> 
> * Auto-close any existing, active body streams/futures on repeated calls to `body.stream`, drop of the `body` resource they originated from or calls to `body.finish`

I fear that this will result in weird spooky-action-at-a-distance situations, and very hard-to-debug issues in code that doesn't expect its handles to be closed behind its back.

Would an alternative be for all of these calls to return results, with an error being returned if the child resource has already been acquired and not closed? That would be similar to what web streams do, for example: if, say, you call `ReadableStream#getReader()`, the stream gets locked, and trying to call `getReader()` again will throw an exception. If you close the reader, you can't use it anymore, but can acquire a new one, because the stream is unlocked.

Note that the only change I'm proposing is to not auto-close existing resources when calling, e.g., `request.body()`. All operations on existing child-resource handles should still have the tombstone semantics, because I agree that auto-closing child handles on moving/closing the parent is much better than trapping.

rvolosatovs:
> > * Auto-close any existing, active body resources on repeated calls to `request.body` or `response.body`
> > * Auto-close any existing, active body streams/futures on repeated calls to `body.stream`, drop of the `body` resource they originated from or calls to `body.finish`
> 
> I fear that this will result in weird spooky-action-at-a-distance situations, and very hard-to-debug issues in code that doesn't expect its handles to be closed behind its back.
> 
> Would an alternative be for all of these calls to return results, with an error being returned if the child resource has already been acquired and not closed? That would be similar to what web streams do, for example: if, say, you call `ReadableStream#getReader()`, the stream gets locked, and trying to call `getReader()` again will throw an exception. If you close the reader, you can't use it anymore, but can acquire a new one, because the stream is unlocked.
> 
> Note that the only change I'm proposing is to not auto-close existing resources when calling, e.g., `request.body()`. All operations on existing child-resource handles should still have the tombstone semantics, because I agree that auto-closing child handles on moving/closing the parent is much better than trapping.

IIRC, the reason we went with auto-close was the `body` accessor on request/response, since only one of them should be able to exist at a time. We should be able to make `body()` return `result<option<body>>`, where the outer result is `err` if `body` is called before previous one has been dropped. The inner option is `none` if the `body` has been `finish`'ed.
We could then "resultify" `body.finish` and `body.stream`.
This gets a bit tricky in the host, because we don't have an explicit `drop` on stream handles, but there are ways to work around that with more involved state tracking.

That actually seems to align with the idea of `body` resource being replaced by `stream<u8, trailers, error-code>`. It seems the most natural for `body` method to return `result<option<stream<u8, trailers, error-code>>>` in that scenario.

Note that I'm opting for `result<option<body>>` rather than `result<body, body-error>`, since getting a "consumed body" is not semantically an error, there's simply _no more data left_, I believe this would also simplify usage in guest languages that don't have rich enum types like e.g. Rust

@lukewagner WDYT?

tschneidereit:
This seems much better to me, thank you! In particular `into-parts` is now much less of a footgun I think: I could very easily see developers using `body()` first, then later deciding they want to get a clone of the headers and using `into-parts()` for that, simply ignoring the body. This pattern won't work any better with this API, but at least it'll fail loudly and immediately, instead of quietly invalidating the other body handle.

lukewagner:
I suppose that's another valid option to consider and I do like the "locality" of the approach.  But I worry that, because GC languages usually lack prompt finalization, an "unreachable but not yet finalized" `Body` object will cause this method to fail and the developer will have no recourse but to somehow find a place where I can call an explicit `.dispose()` method on it before it became unreachable, which sounds like perhaps a worse DX overall.

Taking a step back: if a single component instance calls `request.body` twice in a row, it would be fine if the handle returned by the first call is left working and aliasing the same underlying body as the handle returned by the second call; just like we do with `headers` and just like the browser's JS `fetch()` API does.  The important thing to prevent unintended interference is that, when the parent `request`/`response` is transferred to another component or the host, there are no dangling child `body` handles into it, which is what plain "child" handle semantics gives you.  Thus we could relax the rules on the `body` method to be just like `headers` and avoid both the action-at-a-distance and need-to-explicit-dispose problems.

The only remaining problem is that `into-parts` looks, from a WIT perspective, like "transferring ownership out" and so it "spuriously" triggers the child rules.  Given that one can always pipe a stream from one `request` into a new `body` for a new `request` or `response`, it doesn't seem fundamentally necessary to extract the whole `body` of a `request`/`response` (and indeed, with `stream<u8, trailers, http-error>`, `body` may go away entirely).  Given the special immutability rules for headers though, it *is* necessary to be able to extract the `headers`.  So what if we nixed `into-parts` and instead added a `take-headers: func() -> headers` (and `take-response-options` too)?

rvolosatovs:
If two `body` resources referencing the same request/response could exist at the same time, what would second `finish` do? E.g.:
```rust 
let Some(b1) = req.body();
let Some(b2) = req.body();
let Ok(fut) = b1.finish().await;
let Ok(None) = fut.await;
let x = b2.finish().await; // is this an error?
// If x is an error, how does it differ from
// the case where a reference to the stream exists?
```

Another thing to note is that we'd need to make `headers` method return `option<headers>` and after `take-headers` is called, `headers` would return `none`. Similarly, `take-headers` itself would have to return `option<headers>`

tschneidereit:
I think I agree that we can change things to just return handles representing the same underlying resource, yes. Even better would be if the same handle were returned, but I remember that there are issues with implementing that efficiently.

Regarding `into-parts`: do we really need a replacement for it? We already have `fields#clone()`, and doing `request.headers().clone()` doesn't seem all that onerous to me.

lukewagner:
> If two body resources referencing the same request/response could exist at the same time, what would second `finish` do?

It'd do the same thing as if you called `finish` twice on the same handle (b/c it's the same underlying resource; like how `headers` and mutation works today).

> Even better would be if the same handle were returned, but I remember that there are issues with implementing that efficiently.

Sigh; agreed.  But yeah, some very thorny (perf *and* leak) issues if we try to provide handle identity.

> Regarding into-parts: do we really need a replacement for it? We already have fields#clone(), and doing request.headers().clone() doesn't seem all that onerous to me.

Good point!  That would address Roman's question regarding `headers` needing to return an `option` too.  Hypothetically `take-headers` could be cheaper than COW `clone` on some impls, but I think it's worth starting without it and seeing if anyone complains.

rvolosatovs:
- added `request-options#clone`
- dropped `into-parts`
- made `body()` method return `option<body>` yet again
- documented the fact that duplicate calls to `body.finish`/`body.stream` across instances will cause an error

I think that should address all the feedback, @tschneidereit @lukewagner PTAL

wit-0.3.0-draft/types.wit
```diff
@@ -265,18 +265,30 @@ interface types {
...
+    /// This method returns a stream and a future, which will resolve
     /// to an error code if receiving data from stream fails.
-    /// The returned future resolves to success if body is closed.
+    /// The returned future resolves to success if data section of the
+    /// body is fully consumed.
+    ///
+    /// The handles returned by this method are considered to be child
+    /// handles. Dropping the resource on which this method is called
+    /// will close the handles with an error context, if they are still open.
+    ///
```
rvolosatovs, line 276: Since we can now have multiple live `body` references, do we actually still care about this?

Seems like
```rust
let (stream, fut) = req.body()?.stream()?;
// use stream and fut
drop(stream);
drop(fut)
let (stream, fut) = req.body()?.stream()?;
```

Would be nicer to use. (In Rust, `body` resource returned by `body()` would be dropped immediately after `stream()` returns, right?)

lukewagner: That's a good point!  While we could say that the stream is a child of the *grand*parent, at this point, with all the changes we've made, I'm wondering if we should simply kill the `body` resource entirely.

Incidentally, there is no actual `Body` constructor/type in the Web Fetch API; the [`body`](https://developer.mozilla.org/en-US/docs/Web/API/Response/body) property of `Request`/`Response` returns a `ReadableStream`.  But the Web Fetch API doesn't have trailers, so we have to ask how they work.

So if we use this as our 0.3.x goal state:
```wit
  body: func() -> stream<u8, trailers, error-code>;
```
and work back from there, the question is: how do you abandon the rest of the `u8` stream to skip straight to the `trailers`, like you can do today by calling `finish` before the `%stream` is fully consumed.  I think the answer is another stream built-in (that was in `wasi:io` and we need to add back to achieve parity anyways): `stream.skip`.  In particular, passing a sentinel argument value representing "infinity" to `stream.skip` would mean "skip all the remaining `u8` values, but I do want the `trailers` (or any `error-code` on the way to the `trailers`)".

So if that makes sense as a goal, then perhaps the 0.3.0 stopgap is to add a `skip-contents` method (where I believe "contents" is RFC 9110's name for "the bytes between the headers and trailers") to `request` and `response` and then otherwise do our usual emulation of `stream<T,U,V>` pattern that we do elsewhere?  WDYT?

rvolosatovs: That sounds great to me. This approach would be easier to use and would also make the body state tracking much simpler in the host.

I think we can start with something like:
```wit
/// Returns `none` if another stream is open or stream has been consumed
body: func() -> option<tuple<stream<u8>, future<result<option<trailers>, error-code>>>>;
```

There would currently be no efficient way of skipping the body data to get to the headers, but using the same reasoning as we did in https://github.com/WebAssembly/wasi-http/issues/163#issuecomment-2737704605, I think that's fine for MVP.
Discarding the body data to get to the trailers seems to be a pretty rare use case anyway. Eventually we'd handle it via `stream.skip` (or maybe `stream.drain`?)

Giving up on "incremental changes" here, I'll integrate https://github.com/WebAssembly/wasi-http/pull/165 into this PR, remove `body` resource and proceed with the implementation in https://github.com/bytecodealliance/wasip3-prototyping/pull/58

wit-0.3.0-draft/types.wit
```diff
@@ -230,77 +230,35 @@ interface types {
...
     /// and `authority`, or `headers` which are not permitted to be sent.
     /// It is the obligation of the `handler.handle` implementation
     /// to reject invalid constructions of `request`.
-    constructor(
+    ///
+    /// The returned future resolves to result of transmission of this request.
+    new: static func(
       headers: headers,
-      body: option<body>,
+      contents: stream<u8>,
```
tschneidereit, line 258: This still needs to be optional, since not all requests have bodies, right?

wit-0.3.0-draft/types.wit
```diff
@@ -348,22 +306,28 @@ interface types {
...
+    /// by e.g. `handler.handle`.
+    ///
+    /// This method may be called multiple times.
+    ///
+    /// This method will return `none` if it is called while either:
+    /// - a stream or future returned by a previous call to this method is still open
+    /// - a stream returned by a previous call to this method has reported itself as closed
+    /// Thus there will always be at most one readable stream open for a given body.
+    /// Each subsequent stream picks up where the last stream left off, up until it is finished.
+    body: func() -> option<tuple<stream<u8>, future<result<option<trailers>, error-code>>>>;
```
tschneidereit, line 330: I'm a bit unsure whether using `option` for this makes sense: it seems overloaded for request that just don't have a body, and lead to confusing silent failures.

Given that, ISTM the better, if slightly more onerous, signature would be
```wit
body: func() -> option<result<tuple<stream<u8>, future<result<option<trailers>, error-code>>>, body-in-use-error>>;
```

(Where `body-in-use-error` would be a new thing, where I don't have strong feelings about what it should look like, exactly.)

wit-0.3.0-draft/types.wit
```diff
@@ -408,17 +376,23 @@ interface types {
...
+    /// `trailers` future must resolve to a result.
+    /// If `trailers` resolves to an error, underlying connection
+    /// will be closed immediately.
+    ///
+    /// The returned future resolves to result of transmission of this response.
+    new: static func(
       headers: headers,
-      body: option<body>,
-    );
+      contents: stream<u8>,
```
tschneidereit, line 393: Same comment as for `request`

wit-0.3.0-draft/types.wit
```diff
@@ -427,25 +401,31 @@ interface types {
...
+    /// by e.g. `handler.handle`.
+    ///
+    /// This method may be called multiple times.
+    ///
+    /// This method will return `none` if it is called while either:
+    /// - a stream or future returned by a previous call to this method is still open
+    /// - a stream returned by a previous call to this method has reported itself as closed
+    /// Thus there will always be at most one readable stream open for a given body.
+    /// Each subsequent stream picks up where the last stream left off, up until it is finished.
+    body: func() -> option<tuple<stream<u8>, future<result<option<trailers>, error-code>>>>;
```
tschneidereit, line 429: Same comment as for `request`

tschneidereit (commented):
Overall I like this much better! I left a couple of comments inline that seem relevant, though.

wit-0.3.0-draft/types.wit
rvolosatovs: Previously we discussed with @lukewagner and @dicej that `body` *resource* would be non-optional (#165), that's something that would have addressed https://github.com/WebAssembly/wasi-http/issues/164.

In that scenario requests not carrying a body (like GET) would be represented by empty streams.

Now that there's no `body` resource anymore and it's the request/response constructors that return the transmit future, we can reconsider this, indeed, especially since the host would not be able to do anything with the trailers future in that case anyway.

`body` parameter will have to become an `option<tuple<..>>` though

wit-0.3.0-draft/types.wit
lukewagner: I could be missing it, but I couldn't find any prohibition against trailers in `GET` requests, so if there were an `option` wrapping the `stream<u8>`, I think it'd need to be inside the `tuple`.  However, I haven't been able to find a semantic distinction between `none` and "zero-length contents" inside RFC 9110, so I worry that by introducing this distinction, we might be creating something that content may incorrectly depend on.  (E.g., let's say a component decides to fail if the `option<stream<u8>>` is non-`none`, but now it's rejecting `some` zero-length stream which maybe it gets passed (in a multi-component composition).).

Considering the perf angle (which could be another reason for the `option`):
* As an optimization and convenience, I do think it'd make sense to have the `contents` parameter to `{request,response}.new` be `option<stream<u8>>` as Till suggested, but `none` could be spec-defined to be normalized to an empty stream internally.
* I expect guest code simply won't ask for the content stream for `GET` requests, so source-language stream objects won't be wastefully created.  The CABI overhead of creating closed-on-construction streams should also be light, I think.

So perhaps:
```wit
body: func() -> result<tuple<stream<u8>, future<result<option<trailers>, error-code>>>>;
```
?

wit-0.3.0-draft/types.wit
tschneidereit: Thank you for the explanation; makes sense to me

wit-0.3.0-draft/types.wit
tschneidereit: Good catch: I hadn't considered that my proposal would prevent access to trailers as well.

With that, I agree with the proposed interface, with one slight exception: I do think this is a categorically different error from the ones represented by `error-code`, and that we should make it its own type. Is this perhaps even a pattern that'd make sense to lift up higher and use the same error type for in other places as well?

rvolosatovs: As currently proposed, the error type in this case is actually `unit`, e.g. in Rust it's `()`.
Here's the host implementation with the full Rust type: https://github.com/bytecodealliance/wasip3-prototyping/blob/ecd781215e650dc5eb4e6ec4b2fd60858c084654/crates/wasi-http/src/p3/host/types.rs#L669-L700

Does this address the concern?

tschneidereit: Kinda? It's better than using `error-code` (sorry for missing that it's not), but it does seem very low-effort to define a `body-in-use-error` type and return that instead. And would make this much more self-documenting.

lukewagner: I haven't seen us use that idiom before in WASI of having singleton error variants; there's usually just a `result`.

tschneidereit: oh, you're absolutely right: I had confused myself into thinking that if we move away from `error-code`, we need a specific replacement for it. But since the `result` here is only going to be `err` if the body is already taken out, I agree that that's not needed. Sorry for the noise.

lukewagner (approved):
Ok, looks great with only requested remaining change to put `option` on the content stream in the `new` functions.  Thanks for all the hard work implementing and thinking through all the important error and composition cases @rvolosatovs!

rvolosatovs:
@lukewagner I've just pushed https://github.com/WebAssembly/wasi-http/pull/162/commits/ae895756732e9cd2d62669a999e18dca84e364e5

wit-0.3.0-draft/types.wit
```diff
@@ -230,77 +230,36 @@ interface types {
...
     /// It is the obligation of the `handler.handle` implementation
     /// to reject invalid constructions of `request`.
-    constructor(
+    ///
+    /// The returned future resolves to result of transmission of this request.
+    new: static func(
       headers: headers,
-      body: option<body>,
+      contents: option<stream<u8>>,
+      trailers: future<result<option<trailers>, error-code>>,
```
tschneidereit, line 260: One thought on this: are there downsides to always having to pass a future here? I'm not sure what the overhead (both resource- and boilerplate-wise) of creating an managing a future is. If it's trivial, then this doesn't much matter, but otherwise, should we consider wrapping this into an `option<>`?

wit-0.3.0-draft/types.wit
```diff
@@ -348,22 +307,28 @@ interface types {
...
+    /// by e.g. `handler.handle`.
+    ///
+    /// This method may be called multiple times.
+    ///
+    /// This method will return an error if it is called while either:
+    /// - a stream or future returned by a previous call to this method is still open
+    /// - a stream returned by a previous call to this method has reported itself as closed
+    /// Thus there will always be at most one readable stream open for a given body.
+    /// Each subsequent stream picks up where the last stream left off, up until it is finished.
+    body: func() -> result<tuple<stream<u8>, future<result<option<trailers>, error-code>>>>;
```
tschneidereit, line 331: Depending on the answer to my question above, this might also have to change? Though even if we make the `trailers` ctor parameter optional, I think we could leave this as-is, by constructing a future internally if none was passed.

tschneidereit (commented):
This looks great to me! The one open question (which I unfortunately didn't have earlier, my apologies!) is around the `trailers` parameter to the `request`/`response` constructors—see inline comments.

wit-0.3.0-draft/types.wit
rvolosatovs: In the "stream<T, U, V> world", if we go with optional body in constructors, the parameter would be `option<stream<u8, option<trailers>, error-code>>`, meaning that users could either pass no body stream or pass a body stream, with potentially empty contents, but *mandatory* final element (`option<trailers>` or `error-code`).
From what I understand, there would not be a way to *optionally* pass a final stream element.

With that in mind, if we want to make the future optional, we'd need to take an `option<tuple<stream<u8>, future<result<option<trailers>, error-code>>>>`, which would then directly translate to `option<stream<u8, option<trailers>, error-code>>` in the future. (see also https://github.com/WebAssembly/wasi-http/pull/162#discussion_r2007900713)

FWIW, here's what it takes to create a new future in Rust guest and pass to a constructor: https://github.com/bytecodealliance/wasip3-prototyping/blob/ad4ccf03172b22320c4a22971c51b4f7249a2846/crates/test-programs/src/p3/http.rs#L78-L81
and then send a value: https://github.com/bytecodealliance/wasip3-prototyping/blob/ad4ccf03172b22320c4a22971c51b4f7249a2846/crates/test-programs/src/p3/http.rs#L112
So 2 LoC, plus an import of [`futures::SinkExt`](https://docs.rs/futures/latest/futures/sink/trait.SinkExt.html) to expose [`futures::SinkExt::send`](https://docs.rs/futures/latest/futures/sink/trait.SinkExt.html#method.send)

I would not expect creating a future handle to be an expensive operation, especially considering the fact that we're most likely dealing with network I/O here.

I'm happy to switch to a `tuple` here, if that's what everyone prefers

wit-0.3.0-draft/types.wit
rvolosatovs: I'd rather not introduce a distinction here using the same reasoning as laid out by @lukewagner in https://github.com/WebAssembly/wasi-http/pull/162#discussion_r2007913987

If the `future` was optional, users of this API would have no way of knowing if request/response receipt was successful if the future was `none`.

E.g. if a component read one byte from `stream<u8>`, after which it was closed and `future` was `none` - was the underlying I/O successful or was there a failure?

The `future` resolving to a success means that the incoming request/response has been received it it's entirety. It seems that if it were optional, `none` would be only possible in very niche use cases, e.g. if the resource was constructed by the guest and then requested by itself via `body`.

I would rather avoid having to document/specify this behavior making this interface more self-documenting

rvolosatovs:
Little update:
`wasi:http@0.3.0-draft` as proposed in this PR is implemented in https://github.com/bytecodealliance/wasip3-prototyping/pull/58, with a working round-trip demo available at https://github.com/rvolosatovs/p3-http-test

wit-0.3.0-draft/types.wit
lukewagner: I think the overhead shouldn't be too significant.  Also, it encourages the generally good engineering practice of explicitly signalling the final success or failure of the `request`/`response` in the `result`.

wit-0.3.0-draft/types.wit
lukewagner: Agreed

tschneidereit: thank you for the explanation and references: that all makes sense to me!

tschneidereit (approved):
This turned out great!

lukewagner:
Agreed; great work @rvolosatovs!

PR #165, opened 2025-03-19
feat(p3): model `body` after `stream<u8, trailers, error-code>`

### LINKED ISSUE #163: [v0.3] `handle` implementations cannot forward trailers without trapping while handling transmit errors

### LINKED ISSUE #164: [p3] `handle` implementations have no way of reporting errors occuring after response return

rvolosatovs:
Closes #164 
Closes #163 

As discussed with @dicej and @lukewagner 

- make `body` non-optional for request and response constructors
- take a `future<result<option<trailers>, error-code>>` parameter in `body#new`, remove `new-with-trailers` (refs https://github.com/WebAssembly/component-model/issues/474#issuecomment-2736430031)
- report any transmit errors occurring in the future returned by body constructor

### DIFF
--- a/wit-0.3.0-draft/types.wit
+++ b/wit-0.3.0-draft/types.wit
@@ -240,27 +240,18 @@ interface types {
   /// trailers at any given time.
   resource body {
 
-    /// Construct a new `body` with the specified stream.
+    /// Construct a new `body` with the specified stream and `finish` future.
     ///
-    /// This function returns a future, which will resolve
-    /// to an error code if transmitting stream data fails.
+    /// The `finish` future must resolve to a result once `stream` has finished.
     ///
-    /// The returned future resolves to success once body stream
-    /// is fully transmitted.
+    /// This function returns a future, which will resolve to an error code if
+    /// transmitting of the request or response this body is a part of fails.
+    ///
+    /// The returned future resolves to success once request or response this
+    /// body is a part of is fully transmitted.
     new: static func(
       %stream: stream<u8>,
-    ) -> tuple<body, future<result<_, error-code>>>;
-
-    /// Construct a new `body` with the specified stream and trailers.
-    ///
-    /// This function returns a future, which will resolve
-    /// to an error code if transmitting stream data or trailers fails.
-    ///
-    /// The returned future resolves to success once body stream and trailers
-    /// are fully transmitted.
-    new-with-trailers: static func(
-      %stream: stream<u8>,
-      trailers: future<trailers>
+      finish: future<result<option<trailers>, error-code>>,
     ) -> tuple<body, future<result<_, error-code>>>;
 
     /// Returns the contents of the body, as a stream of bytes.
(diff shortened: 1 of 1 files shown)


rvolosatovs:
integrated into #162, marking as draft

PR #170, opened 2025-07-05
Add `get-*` prefix to getters

badeend:
As proposed in the [last WASI meeting](https://docs.google.com/presentation/d/1on-QuMkOQ-2GCFcJG0DulxKIEDN4KrxveD--gFIzWyQ/edit?usp=sharing), this PR prefixes the following property-like methods with `get-`:
- `request.method`
- `request.path-with-query`
- `request.scheme`
- `request.authority`
- `request-options.connect-timeout`
- `request-options.first-byte-timeout`
- `request-options.between-bytes-timeout`
- `response.status-code`

I don't foresee any trouble with emitting these as properties in source languages.

---

However, the interface also contains:
- `fields.entries`
- `request.options`
- `request.headers` & `response.headers`
- `request.body` & `response.body`

I haven't prefixed these yet. Their names are standalone nouns so my OOP brain wants these to follow the property conventions too. E.g., I'd expect a Java binding generator to produce `getEntries`, `getOptions`, `getHeaders` & `getBody`.
But using the property convention/syntax generally comes with the implicit assumption that accessing it is cheap, idempotent and/or side-effect free. That's not the case here. For example, the following snippet would normally be idiomatic JavaScript code:

```cs
var host = request.headers.entries["Host"];
var forwardedFor = request.headers.entries["X-Forwarded-For"];
```

but actually, this creates two new owned `fields` handles that are never disposed of. And all the `entries` are copied over fully twice as well. So this seemingly fine code is actually inefficient & leaking resources.

I'm ignoring the existence of the `fields.get` method to get the point across :)

---

Ideas on this are welcome. Concretely I want to know how to proceed:
- Should we prefix those methods with `get-*` anyway and leave it for future selves to figure out?
- Should we assume these methods will never become properties and continue this PR as-is?

@lukewagner Thoughts?

### DIFF
--- a/wit-0.3.0-draft/types.wit
+++ b/wit-0.3.0-draft/types.wit
@@ -262,14 +262,14 @@ interface types {
     ) -> tuple<request, future<result<_, error-code>>>;
 
     /// Get the Method for the Request.
-    method: func() -> method;
+    get-method: func() -> method;
     /// Set the Method for the Request. Fails if the string present in a
     /// `method.other` argument is not a syntactically valid method.
     set-method: func(method: method) -> result;
 
     /// Get the combination of the HTTP Path and Query for the Request.  When
     /// `none`, this represents an empty Path and empty Query.
-    path-with-query: func() -> option<string>;
+    get-path-with-query: func() -> option<string>;
     /// Set the combination of the HTTP Path and Query for the Request.  When
     /// `none`, this represents an empty Path and empty Query. Fails is the
     /// string given is not a syntactically valid path and query uri component.
@@ -277,7 +277,7 @@ interface types {
 
     /// Get the HTTP Related Scheme for the Request. When `none`, the
     /// implementation may choose an appropriate default scheme.
-    scheme: func() -> option<scheme>;
+    get-scheme: func() -> option<scheme>;
     /// Set the HTTP Related Scheme for the Request. When `none`, the
     /// implementation may choose an appropriate default scheme. Fails if the
     /// string given is not a syntactically valid uri scheme.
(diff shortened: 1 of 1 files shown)


tschneidereit:
This might be too magical, but one possibility would be to emit bindings that cache returned resource handles for _simple_ `get-` prefixed methods. _Simple_ in this case could mean that the method either returns a handle directly, or maybe also `option<handle-type>`, but not complex things like `result<..>`, such as `body`.

This puts a decent amount of burden on bindings implementers, but that seems maybe manageable?

lukewagner:
@badeend Thanks, lgtm!

> It is not clear to me yet how the bindings can reliably cache properties without the risk of returning stale information. That being said, this is not a problem we need to solve right now, and I do agree that ideally options & headers should be properties.

Incidentally, I was just reading what Firefox does for property caching in WebIDL: there is a [`[Cached]`](https://firefox-source-docs.mozilla.org/dom/webIdlBindings/index.html#cached) annotation that must be accompanied with either [`Pure`](https://firefox-source-docs.mozilla.org/dom/webIdlBindings/index.html#pure) or [`Constant`](https://firefox-source-docs.mozilla.org/dom/webIdlBindings/index.html#constant) which tells the bindings generator and JIT it can cache the return value (in an internal slot of the JS object wrapping the browser's C++ object) (clearing if the setter is called).  Extending to a component setting, we'd also have to flush if there is a setter and a borrowed handle was passed to an import (which the bindings generator could see and perform automatically).  Anyhow, as you said, this can be a later discussion (and maybe caching is too magical?).

